PR #17, opened 2026-05-14
Check running portal backend in doctor output

### LINKED ISSUE #16: Look for xdg-desktop-portal-impl process in doctor output
Currently, `main.py --doctor` shows the folllwing:
```
ok     xdg-desktop-portal     Wayland portal service for KDE/GNOME capture (/usr/lib/xdg-desktop-portal)
  warn   xdg-desktop-portal-kde KDE Wayland portal backend (not found in PATH)
  warn   xdg-desktop-portal-gnome GNOME Wayland portal backend (not found in PATH)
  warn   xdg-desktop-portal-wlr wlroots Wayland portal backend (not found in PATH)
```
Even if no `xdg-desktop-portal` is running (`systemctl --user stop xdg-desktop-portal.service xdg-desktop-portal-whatever`)

It should probably just show a single line with `xdg-desktop-portal-impl` and check if there is one of `xdg-desktop-portal-(kde)|(gnome)|(wlr)` running, which is easily done with ``, can be done with psutil

mzl2233:
This updates the doctor diagnostics to report a single `xdg-desktop-portal-impl` row based on the running portal backend process, rather than listing each backend binary separately. The screen-capture readiness check now also uses that running backend so an installed portal binary is not treated as ready when no desktop portal implementation is active. Verified with `python3 -m compileall diagnostics.py`, targeted diagnostic assertions, and `git diff --check`.

Fixes #16

### DIFF
--- a/diagnostics.py
+++ b/diagnostics.py
@@ -80,6 +80,20 @@ def _first_matching_command(commands: list[str]) -> Optional[str]:
     return None
 
 
+
+def _running_process(names: list[str]) -> Optional[str]:
+    try:
+        result = _run(["ps", "-eo", "comm="], timeout=2.0)
+    except (OSError, subprocess.TimeoutExpired):
+        return None
+    if result.returncode != 0:
+        return None
+    running = set(result.stdout.split())
+    for name in names:
+        if name in running:
+            return name
+    return None
+
 def _python_module_check(module: str, purpose: str, required: bool = False) -> Check:
     if importlib.util.find_spec(module) is not None:
         return Check(module, STATUS_OK, purpose, "python module is installed")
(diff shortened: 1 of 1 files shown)


IlyaP358:
Hello @mzl2233, very thanks for the contribution and for taking the time to look 
into this, the direction is exactly right!

There's one subtle issue with the `ps -eo comm=` approach: on Linux, the 
`comm` field is limited to 15 characters (from `/proc/PID/comm` / 
`/proc/PID/stat`). Since `xdg-desktop-portal-kde` is 22 characters, it 
would always appear as `xdg-desktop-por` in the output and never match, 
so `_running_process` would always return `None`.

The man page for `pgrep` actually [documents](https://man7.org/linux/man-pages/man1/pgrep.1.html) this explicitly:
> "The process name used for matching is limited to the 15 characters 
> present in the output of /proc/pid/stat. Use the -f option to match 
> against the complete command line, /proc/pid/cmdline."

A working implementation using `pgrep -f` (which reads the full, 
untruncated cmdline) i've already finished my fix too and will now push it in **dev**. You can also help with tests on this issue!

Closing this one, but I really appreciate the effort, hope to see more 
contributions from you!

IlyaP358:
Also, if you don't mind, could I include you in the list of contributors on the project's landing page? 
You can say you haven't contributed anything to the project yet, but no, I appreciate any initiative, and you've shown it!

PR #32, opened 2026-05-31
fix: check required GStreamer doctor elements

### LINKED ISSUE #31: Check for pipewiresrc and x264enc in --doctor
Currently, `--doctor` checks if gst-inspect-1.0 is installed, but it doesn't verify if specific required elements like pipewiresrc and x264enc are actually present. On distros like Arch Linux, users can have a completely "ok" doctor output but fail to stream because these GStreamer plugins are missing (#30).

I need to add explicit checks using gst-inspect-1.0 pipewiresrc and gst-inspect-1.0 x264enc.

If pipewiresrc or x264enc are missing on Wayland, show a warn with a hint to install them. The portal backend (GStreamer pipeline) is the only reliable capture path on KDE/GNOME Wayland, wf-recorder is not a valid fallback there.

nanookclaw:
## Summary

Adds explicit `--doctor` diagnostics for the GStreamer elements used by the Wayland portal capture path: `pipewiresrc` and `x264enc`. The new checks use `gst-inspect-1.0 <element>` and report missing elements with distro-oriented install hints for the PipeWire and x264 plugin packages.

I kept these as warning checks so `--doctor` still reports capability gaps without making non-portal environments look hard-failed.

Closes #31

## Verification

- `python -m py_compile src/diagnostics.py`
- `python src/main.py --doctor`
- `python src/main.py --doctor-json`
- focused mocked helper check for present and missing GStreamer elements
- `git diff --check -- src/diagnostics.py`

### DIFF
--- a/src/diagnostics.py
+++ b/src/diagnostics.py
@@ -73,6 +73,47 @@ def _command_check(
     return Check(binary, status, purpose, "not found in PATH")
 
 
+def _gst_element_check(element: str, purpose: str, install_hint: str) -> Check:
+    gst_inspect = shutil.which("gst-inspect-1.0")
+    if not gst_inspect:
+        return Check(
+            f"gst element {element}",
+            STATUS_WARN,
+            f"{purpose} could not be verified",
+            f"gst-inspect-1.0 not found; {install_hint}",
+        )
+
+    try:
+        result = _run([gst_inspect, element], timeout=5.0)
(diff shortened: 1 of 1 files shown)


IlyaP358 (approved):
lgtm 👍

IlyaP358:
Thank you very much for your help!

Issue #29, opened 2026-05-29
Add installation and uninstallation scripts for GitHub users


IlyaP358:
Currently, users who clone the repository directly from GitHub have to manually manage paths, dependencies, and execution. 
focusing on proper packaging (like AUR), having a simple, automated installation script is a great fallback for users on other distributions.

The implementation should include:
`install.sh` and `uninstall.sh`

PR #34, opened 2026-06-01
meta: add install script

alba4k:
Adds an `install.sh` install script, reads `DESTDIR` and `SRCDIR` environnment variables (useful for the PKGBUILD). 

Does _not_ install any documentation, as that should usually be handled by packages. It also does _not_ change the pack in `meta/fluxcast` and `meta/fluxcast.desktop`, those are hard coded. It does _not_ handle dependencies, it just copies files from `$SRCDIR` to `$DESTDIR`

Should be good, please check

Closes https://github.com/IlyaP358/fluxcast/issues/29


alba4k:
I was wondering, should we keep moving our shit to `/opt/fluxcast`, or do you think `/usr/share/fluxcast` would be a better fit?

IlyaP358 (approved):
lgtm 👍

Regarding the paths, I'd leave `/opt/fluxcast` it's purely a matter of taste.
There are also two issues in the readme, but I'll fix them myself in literally two seconds.

Image

After `gtk-update-icon-cache,` the icon isn't loaded, so you need to run the command with explicit paths (under sudo):

`sudo gtk-update-icon-cache /usr/share/icons/hicolor`

Also, a minor detail on line 104: the warning block won't render correctly, but the solution only requires adding one character.


Image

PR #48, opened 2026-06-17
Support Samsung Galaxy Tab S9 FE (1920x1200) and similar minimal-capability WFD sinks

ArulGandhi:
# Support Samsung Galaxy Tab S9 FE (1920x1200) and similar minimal-capability WFD sinks

## Summary

Adds support for streaming at the source monitor's native 1920x1200 resolution
to Samsung Galaxy Tab S9 FE (and likely other Samsung tablets / Android sinks
that advertise minimal WFD capabilities in M3 but accept higher modes in
practice). Tested on Hyprland + `wf-recorder` backend at 1200p60.

Behaviour before this PR:

```
[FluxCast WFD RTSP]   wfd_video_formats: 00 00 01 01 00000001 00000000 ... none none
[FluxCast WFD RTSP] ERROR: The sink did not advertise a supported 720p/1080p
                    CEA WFD mode. CEA mask was 0x00000001, H.264 level was 01.
```

Session aborts; nothing streams.

Behaviour after this PR:

```
[FluxCast WFD RTSP] WARNING: Sink lacks advertised support for 1920x1200p60;
                    forcing it (most sinks accept it).
[FluxCast WFD RTSP] Negotiated media mode: 1920x1200p60
[FluxCast WFD RTSP] Selected video format: 00 00 01 40 00000000 20000000 ...
[FluxCast WFD RTSP] Starting media as 1920x1200p60; RTP source port 19002
[FluxCast WFD Media] Capturing screen : eDP-1 (1920x1200)
                    ... stream runs at native 1200p60 with no black bars.
```

### 1. `wfd: use ultrafast preset and raise bitrate floors for high-res 60fps`

* `wf-recorder` and `x11grab` ffmpeg pipelines were using `-preset veryfast`
  while the test-pattern and GStreamer portal pipelines already used
  `ultrafast` / `speed-preset=ultrafast`. Align them. `veryfast` can't keep up
  with 1080p60+ encoding in real time on typical laptop hardware.
* Raise `_quality_floor_kbits` for high-resolution 60fps:
    * 1080p60: 12 Mbps → 14 Mbps
    * `>1080p` 30fps: 12 Mbps → 14 Mbps
    * `>1080p` 60fps: 16 Mbps → 20 Mbps

  `ultrafast` has weaker compression efficiency than `veryfast`, so the floor
  needs to be higher to keep visible quality the same.

### 2. `wfd: support 1920x1200 VESA sinks and tolerate minimal CEA advertisements`

Three coupled changes that all live inside `_choose_cea_mode` /
`_selected_video_format`, so they're committed together (splitting them
further leaves an intermediate state that doesn't import).

#### a) VESA 1920x1200p30 / p60 modes

The Wi-Fi Display Technical Specification carries 1920x1200 in the **VESA**
resolution bitmask (bits 28 and 29 — confirmed against the AOSP
[`VideoFormats.cpp`](https://android.googlesource.com/platform/frameworks/av/+/1b92868/media/libstagefright/wifi-display/VideoFormats.cpp)
resolution table), not the CEA bitmask. The previous codebase only modelled
CEA modes, so 1200p was unreachable.

`WFDCEAMode` gains a `table: str = "cea"` field (`"cea"` or `"vesa"`).
`_choose_cea_mode` now selects across the union of `WFD_CEA_MODES` and
`WFD_VESA_MODES`, preferring VESA 1200p when the source monitor is
≥1920×>1080. `_selected_video_format` places the chosen mode bit in the
correct M4 mask field, so the sink allocates a 1920x1200 viewport instead of
letterboxing 1080p (which produced visible black bars during initial
investigation).

> The class name `WFDCEAMode` is now mildly misleading (it also represents
> VESA modes). I left it alone to keep the diff small; happy to rename to
> `WFDDisplayMode` in a follow-up if you prefer.

#### b) Permissive fallback when sink advertises a minimal CEA mask

The Samsung Tab S9 FE responds to M3 with literally:

```
wfd_video_formats: 00 00 01 01 00000001 00000000 00000000 00 0000 0000 00 none none
```

That's CEA bit 0 (640x480p60) and H.264 level 3.1, with no VESA support
advertised — despite the device happily decoding 1080p / 1200p from Windows
Miracast. The previous code raised `WFDNotReady` and bailed.

Instead of failing, log a one-shot warning and force the best mode for the
source monitor:

```python
[FluxCast WFD RTSP] WARNING: Sink lacks advertised support for 1920x1200p60;
                    forcing it (most sinks accept it).
```

This mirrors what Windows Miracast does in practice: the M3-declared CEA / VESA
masks are treated as a hint, and higher-resolution H.264 streams are accepted
by the sink regardless. The warning is deduplicated via a module-level
`_mode_force_warned` flag because `_choose_cea_mode` is called several times
per session (M3 response handling, M4 request build, `_start_media`, and the
active-probe path).

The bit-0 constant `WFD_CEA_640P60` is added so the lookup table is complete
on the parse side.

#### c) H.264 level 5.0 / 5.1 for resolutions exceeding 1920x1080

1920x1200@60 doesn't fit H.264 level 4.2:

* Frame size: `120×75 = 9000 MBs` > level 4.2 limit (8704 MBs)
* Frame rate: `540000 MB/s` > level 4.2 limit (522240)

`libx264` emits warnings like `frame MB size (120x75) > level limit (8704)`
which some sinks then reject or mis-decode. Both `_wfd_level_for_mode`
(M4 bitmask) and `_h264_level_for_mode` (ffmpeg `-level:v`) now return 5.0
(30 fps) / 5.1 (60 fps) when the chosen mode exceeds 1080p. New
`WFD_LEVEL_50` / `WFD_LEVEL_51` constants are added for the M4 bitmask side.

## Testing

* **Hardware:** Intel laptop, eDP-1 = 1920x1200, Samsung Galaxy Tab S9 FE.
* **Compositor:** Hyprland (wlroots).
* **Backend:** auto-selected `wf-recorder` → `ffmpeg` → `rtp_mpegts`.
* **Connection:** NetworkManager Wi-Fi Direct, P2P group owner role on the tab.
* **Result:** Full 1920x1200 stream at 60 fps, no black bars, no encoder
  warnings, no `WFDNotReady` failure. Latency is roughly on par with what
  Windows Miracast delivers to the same tablet (≈1–2 s end-to-end, dominated
  by the Android sink's decode buffer).

Not retested on the LG TV path that already worked. The is-LG branches in
`_vbv_bufsize` and the GStreamer portal pipeline are untouched, and the new
fallback path is gated on the sink not advertising any 720p/1080p mode —
LG TVs advertise CEA correctly, so this code does not affect them.

## Things this PR deliberately does not do

* It does **not** implement `wfd_preferred_display_mode` (WFD spec §6.1.14),
  which is the spec-clean way to negotiate arbitrary timings. The Samsung
  tablet does not advertise `preferred-display-mode-supported=01` in M3, so
  the source can't use that path against this sink anyway. The forced-mode
  approach is what's actually interoperable.
* It does **not** add `wfdx_video_formats` (the Microsoft extension). Same
  reason — the tablet doesn't advertise it.
* It does **not** add HDCP. The Samsung tablet shows a "connection is not
  secure, cannot show protected content" toast on session start; streams work
  fine without HDCP.

## Risks / things to look at

* `WFDCEAMode.table` is a new field with a default of `"cea"`, so any external
  call site that constructs a `WFDCEAMode` directly will still produce a CEA
  mode (i.e. preserves old behaviour). Grepped the tree: only the module-
  internal tables construct it.
* The forced-mode warning fires every time someone re-runs the source against
  a minimal-capability sink. If you'd prefer it gated behind a CLI flag (e.g.
  `--allow-non-advertised-modes`) that's a 5-line change — happy to do that
  instead.

### DIFF
--- a/src/wfd.py
+++ b/src/wfd.py
@@ -24,14 +24,22 @@ except OSError:
 
 # WFD CEA resolution bitmask. The current backend intentionally negotiates
 # only the !common! HD modes that Samsung TVs usually accept reliably.
-WFD_CEA_720P30 = 0x00000020   # bit 5: 1280x720p30, mandatory HD mode
-WFD_CEA_720P60 = 0x00000040   # bit 6: 1280x720p60
+WFD_CEA_640P60  = 0x00000001  # bit 0: 640x480p60
+WFD_CEA_720P30  = 0x00000020  # bit 5: 1280x720p30, mandatory HD mode
+WFD_CEA_720P60  = 0x00000040  # bit 6: 1280x720p60
 WFD_CEA_1080P30 = 0x00000080  # bit 7: 1920x1080p30
 WFD_CEA_1080P60 = 0x00000100  # bit 8: 1920x1080p60
+
+# VESA resolution bitmasks (Table 5-11 / AOSP VideoFormats.cpp)
+WFD_VESA_1200P30 = 0x10000000  # bit 28: 1920x1200p30
+WFD_VESA_1200P60 = 0x20000000  # bit 29: 1920x1200p60
+
 WFD_LEVEL_31 = 0x01
 WFD_LEVEL_32 = 0x02
 WFD_LEVEL_40 = 0x04
 WFD_LEVEL_42 = 0x10
+WFD_LEVEL_50 = 0x20
+WFD_LEVEL_51 = 0x40
 WFD_AUDIO_AAC = "AAC 00000001 00"
 NM_DEST = "org.freedesktop.NetworkManager"
 NM_PATH = "/org/freedesktop/NetworkManager"
@@ -171,6 +179,7 @@ class WFDCEAMode:
     width: int
     height: int
     fps: int
+    table: str = "cea"  # "cea" or "vesa"
 
     @property
     def resolution(self) -> str:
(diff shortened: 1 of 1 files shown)


IlyaP358 (approved):
lgtm👍. Changing the base branch to **dev** to merge it there first. I need to update the AppImage version and PyPI package version before pushing the final release to main. Thanks for the contribution!

PR #66, opened 2026-07-09
Add per-mode tray configuration

### LINKED ISSUE #65: Tray: per-mode config file so users can override stream defaults without the CLI
The tray always launches sessions on the built-in defaults (fps=30, bitrate=4M, auto resolution). To change resolution / fps / bitrate today you have to drop to the CLI and learn the flags. Tray users have no way to, say, run WFD at 60fps or set a higher bitrate.

Proposal
A user-editable config file, read only by the tray, that fills in stream flags per mode. The CLI (python3 src/main.py …) stays completely untouched.

Location: $XDG_CONFIG_HOME/fluxcast/config (fallback ~/.config/fluxcast/config). A missing/empty file means today's exact behavior.
Format: INI (configparser, stdlib, already used in tray.py). No new dependencies.
Sections = the three tray modes: [wfd], [dlna], [cast]. The tray reads only the section for the mode being launched.
Keys = the exact CLI long-flag names without -- (output-res, fps, bitrate, wfd-uibc, …). No mental translation.
How it behaves (precedence)
For a tray launch: config value overrides the built-in default; any key not in the config keeps its default. Per-mode duplication is intentional (e.g. a different bitrate per mode).

Example - set fps to 60 for WFD but leave resolution on its default:

`# ~/.config/fluxcast/config`

```
[wfd]
output-res = 1920x1080
fps = 60
bitrate = 8M
wfd-uibc = true

[dlna]
transport = hls
fps = 30
bitrate = 4M

[cast]
bitrate = 4M
```
Here WFD tray runs get --output-res 1920x1080 --fps 60 --bitrate 8M --wfd-uibc; DLNA gets --transport hls --fps 30 --bitrate 4M; anything omitted stays default.

Implementation sketch
New src/config.py with something like load_profile(mode) -> list[str] returning the extra argv for that mode's section (["--fps", "60", …]), or [] if no file/section.
tray.py: after building the base cmd, append load_profile("wfd"|"dlna"|"cast").
store_true flags (wfd-uibc, wfd-no-audio): = true appends the flag, anything else / absent skips it.
Validate keys against a known-flags whitelist per mode; unknown keys -> warn to the tray log and skip (don't crash, don't blindly forward arbitrary args).
Invalid values -> warn + fall back to default.

Yurii201811:
## Summary

Adds an XDG-compliant, per-mode INI configuration file for tray launches. Tray users can override stream defaults without changing direct CLI behavior, while invalid or unknown options are logged and ignored. Missing files, empty files, and omitted keys preserve today’s defaults.

## Background

The tray currently always launches WFD, DLNA, and Cast with built-in defaults, so users must switch to the CLI to change resolution, FPS, or bitrate.

Closes #65

## Changes

- add `src/tray_config.py` with per-mode whitelists, value validation, boolean flag handling, and safe warnings
- append validated profile arguments to WFD, DLNA, and Cast tray commands
- add seven regression tests for defaults, mode isolation, malformed input, validation, and XDG path resolution
- document the config path, INI format, accepted keys, and fallback behavior

## Testing

- `python3 -m unittest tests.test_tray_config -v` — 7 passed
- `python3 -m compileall -q src tests` — passed
- `git diff --check origin/dev...HEAD` — passed
- `python3 -m unittest discover -s tests -v` — 15 passed; 2 unchanged firewall diagnostics tests fail on macOS because their mocks expect Linux firewalld ordering

## Additional Notes

The whitelist follows the CLI flags currently available on `dev`. The issue example mentions `wfd-uibc`, but that flag has not landed on `dev`; it is therefore reported as unknown instead of being forwarded to an argparse version that would reject it.

### DIFF
--- a/documentation/DOCUMENTATION.md
+++ b/documentation/DOCUMENTATION.md
@@ -92,6 +92,51 @@ python3 src/main.py --protocol cast
   - Requires `libappindicator` (Hyprland/KDE) or `gnome-shell-extension-appindicator` (GNOME).
   - On non-Hyprland Wayland (KDE, GNOME), WFD capture uses the xdg-desktop-portal screen picker dialog.
 
+#### Per-mode tray configuration
+
+Tray launches can override stream defaults with an INI file at
+`$XDG_CONFIG_HOME/fluxcast/config`, or `~/.config/fluxcast/config` when
+`XDG_CONFIG_HOME` is not set. This file affects only the tray; direct CLI
+commands keep their normal behavior.
+
+```ini
+[wfd]
+output-res = 1920x1080
+fps = 60
+bitrate = 8M
+wfd-no-audio = false
+
+[dlna]
+transport = hls
+fps = 30
+bitrate = 4M
+
+[cast]
+bitrate = 4M
+```
+
+All modes accept `output-res`, `fps`, and `bitrate`. The `dlna` and `cast`
+sections also accept `host`, `port`, `discover-timeout`, `transport`, and
+`capture-backend`. The `wfd` section accepts these WFD stream/session options:
+
+- `wfd-test-pattern`
+- `wfd-media-pipeline`
+- `wfd-capture-backend`
+- `wfd-latency-log`
+- `wfd-no-audio`
+- `wfd-audio-device`
+- `wfd-rtsp-port`
+- `wfd-no-firewall`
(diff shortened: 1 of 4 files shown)


IlyaP358 (approved):
lgtm! thanks.

PR #70, opened 2026-07-13
feat(doctor): warn when an interface overlaps the Wi-Fi Direct subnet

### LINKED ISSUE #68: Update --doctor to check for network conflicts
Add an option so --doctor will also check for conflicts with 192.168.49.1/24 since it's the hardcoded default IP address that NetworkManager uses to provision Wi-Fi Direct (P2P) groups.

ravi-arnan:
Closes #68

Adds a read-only \`_subnet_conflict_check()\` to \`--doctor\` that enumerates interface addresses via \`ip -j addr\` and warns on any overlap with the NetworkManager Wi-Fi Direct subnet 192.168.49.0/24, skipping \`p2p-*\` interfaces (which live there by design). Includes unit tests and a live \`--doctor\` run.

Note: the existing FirewallCheckTest has 2 failures on \`dev\` unrelated to this change (stale tests vs the current \`_firewalld_check\`), happy to address separately if useful.

### DIFF
--- a/src/diagnostics.py
+++ b/src/diagnostics.py
@@ -1,5 +1,6 @@
 import json
 import importlib.util
+import ipaddress
 import os
 import platform
 import re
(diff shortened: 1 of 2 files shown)


IlyaP358 (approved):
lgtm👍 Thanks for the PR. If you want to open another to fix those firewall tests, I'd be happy to review it. =]

PR #71, opened 2026-07-14
fix(doctor): probe the firewalld port instead of always warning

ravi-arnan:
Follow-up to #70, as you offered in review. =]

The `--doctor` firewalld check had two red tests in `tests/test_diagnostics.py`:

- `test_firewalld_running_without_port_warns`
- `test_returns_worst_case_across_front_ends`

**Root cause:** `_firewalld_check()` warned whenever firewalld was *running*, without ever checking whether the WFD RTSP port (`7236/tcp`) was actually open, and it read daemon state via `systemctl is-active`. The tests expect it to query firewalld directly and report OK when the port is already allowed.

**Fix:** query firewalld with `firewall-cmd --state` and `firewall-cmd --query-port=7236/tcp`:

- not running -> **OK** (port not blocked)
- running, port open -> **OK**
- running, port closed -> **WARN** (with the `--add-port` remediation hint)

This mirrors the existing `ufw` path, which only warns when the port is genuinely closed. The old code carried a comment worrying that probing could prompt for auth — that only applies to `--add-port`; `--query-port` is a read-only D-Bus getter and never prompts.

All 17 firewall/subnet tests pass (`python3 -m unittest tests.test_diagnostics`); full suite green (27 tests).

### DIFF
--- a/src/diagnostics.py
+++ b/src/diagnostics.py
@@ -519,25 +519,37 @@ def _firewalld_check() -> Optional[Check]:
         return None
 
     try:
-        state = _run(["systemctl", "is-active", "firewalld"], timeout=3.0)
+        state = _run(["firewall-cmd", "--state"], timeout=3.0)
     except (OSError, subprocess.TimeoutExpired) as exc:
         return Check("firewall (firewalld)", STATUS_WARN, "could not query firewalld state", str(exc))
 
-    if state.stdout.strip() != "active":
+    if state.returncode != 0 or state.stdout.strip() != "running":
         return Check(
             "firewall (firewalld)",
             STATUS_OK,
             "firewalld is not running; port not blocked",
-            state.stdout.strip(),
+            (state.stdout + state.stderr).strip(),
         )
 
-    # Don't probe the port here: `firewall-cmd --query-port` can prompt for auth, and
-    # FluxCast opens the port automatically for the session anyway.
+    # `--query-port` is a read-only D-Bus getter; unlike `--add-port` it never
+    # prompts for authentication, so probing the port here is safe.
+    try:
+        query = _run(["firewall-cmd", f"--query-port={WFD_RTSP_PORT}/tcp"], timeout=3.0)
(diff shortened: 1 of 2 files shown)


IlyaP358 (changes requested):
Thanks, the direction is right (OK when the port is actually open beats always warning).

One blocker though: `--state`/`--query-port` aren't prompt-free, firewalld gates them through polkit (dialog on desktop, "Authorization failed" over SSH). And there's a real bug from that: in the `--state` branch a non-zero auth failure is read as "firewalld not running -> OK", so on a gated/headless box doctor gives a false all-clear. The `--query-port` branch has the same issue (auth error shows up as "port closed").

Fix would be to only trust what firewall-cmd literally prints: treat firewalld as stopped only when it actually says "not running", and call the port open or closed only on a real "yes" or "no". For anything else, like an auth error or unexpected output, just warn that you couldn't verify it instead of assuming it's fine or that it's closed. That way an auth failure never turns into a definitive result either way. Happy to help if useful

ravi-arnan:
Good catch on the polkit gating — you're right that a non-zero exit was being read as a definitive result, so an auth failure gave a false all-clear (and the port branch a false "closed"). Fixed in 6acda1a.

The check now trusts only what `firewall-cmd` literally prints:
- firewalld is treated as stopped **only** when the output contains `not running`
- the port is called open/closed **only** on a literal `yes`/`no`
- anything else — an auth error, empty or unexpected output — now WARNs that it *couldn't verify* the state/port, instead of assuming OK or closed

So an `Authorization failed.` never turns into a definitive result either way. Added two regression tests covering the auth-failure path on both `--state` and `--query-port`; all 19 diagnostics tests pass. Thanks for the thorough review!

IlyaP358 (approved):
Verified locally: the auth-failure path now returns a neutral "couldn't verify" WARN on both --state and --query-port instead of a false OK/closed, and the 19 tests (incl. your two new ones) pass. Both polkit outcomes degrade cleanly now. LGTM, merging, thanks for the quick and thorough fix!

PR #75, opened 2026-07-18
docs: document --wfd-go-intent and MAC tip for --wfd-peer

### LINKED ISSUE #74: Document --wfd-go-intent and add a --wfd-peer MAC tip
Two WFD changes landed on dev recently and still need documenting.

1. Add `--wfd-go-intent` (0-15, default 0). Currently undocumented. Add it in two places:
  The WFD Flags list, after --wfd-timeout.
  The WFD Discovery/Connect section, with a one-line description.
  
For the description, do not state that it "makes the TV the group owner." A tester confirmed the fix works while the P2P address stayed on 10.42.0.x, so the sink is not actually taking the owner role. Describe it accurately instead: the flag sets FluxCast's Wi-Fi Direct group-owner intent, and a low value (0 by default) is what gets most Miracast TVs to start the session. A higher value can be used if a specific sink requires it. No claims about the IP range changing.

2. Update `--wfd-peer`. Add a note recommending selection by MAC rather than index, since the index can change between a separate --wfd-scan and the connect step.

No code changes required.

gitcommit90:
## Summary
Documents two recent Wi-Fi Direct CLI surfaces that were missing from `DOCUMENTATION.md`, per #74:

1. `--wfd-go-intent` (`0`–`15`, default `0`) with accurate GO-intent wording
2. Tip to prefer MAC over index for `--wfd-peer` (indices can change between scan and connect)

## Changes
- List `--wfd-go-intent 0-15` under **WFD Flags**
- Document the flag under **WFD Discovery/Connect** without claiming the TV becomes GO or that the P2P address range changes
- Note MAC preference under `--wfd-peer`

Docs-only; no code changes.

## Tests
```
python3 -m pytest tests/ -q
29 passed in 0.09s
```

Closes #74

### DIFF
--- a/documentation/DOCUMENTATION.md
+++ b/documentation/DOCUMENTATION.md
@@ -67,6 +67,7 @@ python3 src/main.py --protocol cast
 - `--wfd-no-firewall`
 - `--wfd-interface IFACE`
 - `--wfd-timeout SEC`
+- `--wfd-go-intent 0-15`
 - `--wfd-monitor NAME` **deprecated** alias for `--monitor` (kept for compatibility)
 
 ## Flag Details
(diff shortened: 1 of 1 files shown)


IlyaP358 (approved):
lgtm👍. Thanks a lot of!

PR #77, opened 2026-07-22
Avoid redundant firewalld authorization prompts

### LINKED ISSUE #76: firewalld: check the port before the privileged --add-port to avoid a needless Polkit prompt
Reported on #72 (https://github.com/IlyaP358/fluxcast/issues/72#issuecomment-5038378541): on LG OLED55B87LC the WFD session works, but the automatic firewalld handling triggers two Polkit password dialogs and then prints Could not open firewalld port 7236/tcp automatically, even though 7236/tcp is already open in the nm-shared zone. `--wfd-no-firewall` avoids it.

`_open_wfd_firewall_port` (in src/wfd.py) goes straight to the privileged `firewall-cmd --zone=nm-shared --add-port=<port>/tcp`. It only notices `ALREADY_ENABLED` after the call, so the Polkit prompt has already fired even when nothing needs doing.

Probably fix: before the privileged add, run `firewall-cmd --zone=nm-shared --query-port=<port>/tcp`. If it returns `yes`, skip the add (no prompt, no error). If the query itself fails or times out (on strict polkit setups even reads need auth - see note), don't fall through to `--add-port`; bail out with the `--wfd-no-firewall` hint instead of firing more prompts. Add only when the port is confirmed closed.

Note: runtime state confirmed on the thread, the port is open in runtime. On the reporter's system polkit requires auth even for read-only `firewall-cmd ` queries (only` --state` is exempt), so the pre-query prompts there too and won't avoid the dialog; that's the "two prompts then timeout / Did not receive a reply" case. The pre-query fixes the common case (reads unprivileged); strict-polkit systems rely on the bail-out branch above + `--wfd-no-firewall`.

donglrd:
## Summary

- query the `nm-shared` firewalld zone before requesting a privileged port add
- leave an already-open port untouched
- fail closed with the existing manual-firewall hint when the query cannot be verified
- add unit coverage for the open, closed, authorization-error, and timeout paths

Fixes #76

## Testing

- `PYTHONPATH=src python3 -m unittest discover -s tests -q` (33 tests)
- `python3 -m compileall -q src tests`

The query uses firewalld's normal `yes`/`no` exit/status pair; errors and timeouts do not fall through to `--add-port`.

### DIFF
--- a/src/wfd.py
+++ b/src/wfd.py
@@ -2392,6 +2392,7 @@ def _firewalld_active() -> bool:
 
 
 _FIREWALL_AUTH_TIMEOUT = 60.0
+_FIREWALL_QUERY_TIMEOUT = 3.0
 
 _WFD_FIREWALL_ZONE = "nm-shared"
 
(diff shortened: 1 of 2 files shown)


IlyaP358 (approved):
lgtm👍. thank you very much for the pr =]

Issue #81, opened 2026-07-25
No way to see ffmpeg progress stats (fps, dup, drop)


IlyaP358:
The ffmpeg senders run with `-loglevel warning` and no `-stats`, so the progress line is suppressed and only occasional warnings like `More than 1000 frames duplicated` make it into the output. That means testers can't report the actual capture rate, which is exactly what's needed to tell a capture problem from a link problem.

Came up while debugging a stuttering report where we had no way to know whether x11grab was keeping up. Adding `-stats` (or a debug/verbose flag that enables it) would make these reports diagnosable.

PR #83, opened 2026-07-26
fix(wfd): show ffmpeg progress stats

w3lld1:
## Summary

- I enabled ffmpeg's progress line while keeping the existing warning log level.
- I applied the shared sender arguments to test-pattern, wf-recorder, and x11grab WFD streams.
- I added a regression test for the progress arguments.

## Validation

- `python3 -m unittest tests.test_wfd.FfmpegProgressArgsTest -v`
- `python3 -m unittest discover -s tests` (46 tests)
- `python3 -m compileall -q src tests`
- `git diff --check origin/dev...HEAD`

Fixes https://github.com/IlyaP358/fluxcast/issues/81

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -36,6 +36,7 @@ Wi-Fi Display (Miracast) Options:
     --wfd-peer PEER          Peer selector: index, MAC, or name
     --wfd-dry-run            Print D-Bus call without activating connection
     --wfd-test-pattern       Stream generated test video instead of the desktop
+    --wfd-ffmpeg-stats       Show ffmpeg progress statistics for WFD streams
     --wfd-media-pipeline auto|ffmpeg|gst
                              RTP media sender (default: auto)
     --wfd-capture-backend auto|portal|wf-recorder|x11grab
@@ -133,6 +134,8 @@ def parse_args() -> argparse.Namespace:
                      help="Print WFD connection D-Bus call without activating it")
     wfd.add_argument("--wfd-test-pattern", action="store_true", dest="wfd_test_pattern",
                      help="For --protocol wfd, stream generated test video instead of the desktop")
+    wfd.add_argument("--wfd-ffmpeg-stats", action="store_true", dest="wfd_ffmpeg_stats",
+                     help="For --protocol wfd, show ffmpeg progress statistics")
     wfd.add_argument("--wfd-media-pipeline", default="auto",
                      choices=["auto", "ffmpeg", "gst"],
                      dest="wfd_media_pipeline",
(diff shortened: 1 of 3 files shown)


IlyaP358 (changes requested):
Hi, thanks for the pr. The approach is right and the refactor into a shared helper is clean. I confirmed `-loglevel warning -stats` does surface the progress line, and all three ffmpeg paths are covered.

One thing before merging. ffmpeg refreshes that line about twice a second with a carriage return, and it is longer than most of FluxCast's own output, so leftovers get glued straight onto our messages. A normal session renders like this:

```
[FluxCast WFD Media] Sender health: pid=5341:running; tx+2967340 KiB on p2p-wlan0-0d=0:00:01.50
[FluxCast WFD RTSP] M16 keepalive sentme=00:00:04.50 bitrate=N/A speed=1.12x elapsed=0:00:04.00
[FluxCast WFD RTSP] <- response: RTSP/1.0 200 OK5.00 bitrate=N/A speed=1.11x elapsed=0:00:04.50
```

The interface reads as `p2p-wlan0-0d`, the status as `200 OK5.00`. Nothing is lost, but it looks broken, and it ends up in the logs testers paste into issues.

My issue text did say "-stats or a debug/verbose flag", so that's on my wording, not on you. Could you put it behind a flag instead?

- add `--wfd-ffmpeg-stats` in `src/main.py` (store_true, plus a usage line)
- add `ffmpeg_stats: bool = False` to `WFDMediaConfig`
- pass it at the single config build site, `src/wfd.py:3534`
- make the helper take it, `_ffmpeg_sender_args(show_stats)`, appending `-stats` only when set
- update the three call sites

Without the flag the args stay exactly as today, so nothing changes for existing users. The rest of the PR is good as is.

Optional: the test asserts the helper returns its own literal list. Checking that a built command contains `-stats` would prove more.

w3lld1:
Thanks — addressed this in `f2909d6`:

- added opt-in `--wfd-ffmpeg-stats` CLI flag;
- kept the default ffmpeg arguments unchanged, without `-stats`;
- threaded the setting through `WFDMediaConfig` to all three ffmpeg sender paths;
- covered both the default and enabled argument behavior.

Fresh validation:
- `python3 -m unittest tests.test_wfd` — 6 passed
- CLI help assertion for `--wfd-ffmpeg-stats`
- `git diff --check`

IlyaP358 (approved):
Perfect, this is exactly what I had in mind. Verified the default path produces byte-identical ffmpeg commands to dev, so nothing changes for existing users, and the flag turns stats on across all three senders. Thanks for the quick turnaround, and for tightening the tests too. Merging =]

PR #88, opened 2026-07-31
Skip firewall diagnostics with --wfd-no-firewall

### LINKED ISSUE #82: --wfd-no-firewall still triggers a Polkit prompt from the doctor
Reported by @Bibergoahead on #79. Starting FluxCast with `--wfd-no-firewall` still pops a Polkit password dialog during the doctor phase, right before the monitor prompt. It goes away after ~3 seconds on its own, and the doctor then reports a firewalld timeout.

Not a duplicate of #76: that one was the runtime port add in `_open_wfd_firewall_port`. This is the doctor's own check. `run_diagnostics()` runs unconditionally in `start_experimental_backend`, and `_firewalld_check()` in `src/diagnostics.py` doesn't know about the flag, so it calls `firewall-cmd --state` regardless.

Two things go wrong because of it: the 3 second timeout is added to every startup on such systems, and the dialog steals terminal input, so the monitor picker gets a stray keypress and falls back with `Invalid choice, using monitor 0`.

If the user explicitly said not to touch the firewall, the firewalld check should be skipped or reported as `skip`.

lntutor:
## Summary

Honor `--wfd-no-firewall` during the startup doctor phase by forwarding the flag to diagnostics and returning a skipped firewall check without invoking `ufw` or `firewall-cmd`.

This avoids the Polkit prompt and timeout before monitor selection while keeping the rest of the readiness report unchanged.

## Verification

- `python3 -m unittest discover -s tests` (49 tests)
- `python3 -m compileall -q src`

Fixes #82

### DIFF
--- a/src/diagnostics.py
+++ b/src/diagnostics.py
@@ -662,7 +662,12 @@ def _python_check() -> Check:
     )
 
 
-def run_diagnostics() -> DiagnosticReport:
+def run_diagnostics(skip_firewall: bool = False) -> DiagnosticReport:
+    firewall_check = (
+        Check("firewall", STATUS_SKIP, "disabled by --wfd-no-firewall")
+        if skip_firewall
+        else _firewall_check()
+    )
     checks = [
         _python_check(),
         _command_check("ffmpeg", "video/audio transcoding", required=True),
@@ -702,7 +707,7 @@ def run_diagnostics() -> DiagnosticReport:
         _iw_p2p_check(),
         _supplicant_capability_check(),
         _supplicant_wfd_check(),
-        _firewall_check(),
+        firewall_check,
     ]
 
     by_name = {check.name: check for check in checks}
--- a/src/main.py
+++ b/src/main.py
@@ -244,7 +244,7 @@ def main() -> None:
 
     if args.doctor or args.doctor_json:
         from diagnostics import print_report, run_diagnostics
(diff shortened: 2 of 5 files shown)


IlyaP358 (changes requested):
Nice, clean fix.

One thing before merge, if you're up for it: `--doctor` and `--doctor-json` still call `run_diagnostics()` bare in `src/main.py`, so `--doctor --wfd-no-firewall` keeps the prompt and the 3s wait. Same one-liner.

lntutor:
Addressed on new head `0f29469`.

`--doctor` and `--doctor-json` now pass `--wfd-no-firewall` through `run_diagnostics(skip_firewall=...)`, so the diagnostics path no longer prompts/waits on the firewall check when the flag is set.

Verification on this exact head:
- `python3 -m unittest tests.test_diagnostics -v`
- `python3 -m unittest tests.test_wfd.StartBackendDiagnosticsTest.test_no_firewall_flag_is_forwarded_to_diagnostics -v`
- command-level check: `python3 src/main.py --doctor-json --wfd-no-firewall` reports the `firewall` check as `skip` with message `disabled by --wfd-no-firewall`.

Please re-review when convenient.

IlyaP358 (approved):
Yep. Thank you. merging =]

PR #91, opened 2026-08-22
Fix P2P neighbor selection for active probes

### LINKED ISSUE #85: Active RTSP probe can target the TV's home-network address instead of its P2P address
Reported by @Bibergoahead in #79. When a TV is connected both to the home network and to our P2P group, and it uses the same MAC on both, the active probe can pick the wrong address:

```
[FluxCast WFD RTSP] Active probe: TV=192.168.178.54; connecting to RTSP port 7236...
```

while `ip neigh` had the right one all along:

```
10.42.0.182      dev p2p-wlan0-3
192.168.178.54   dev wlan0
```

Cause: `_get_peer_ip_from_arp` in `src/wfd.py` scans the whole neighbour table and matches on MAC with no interface filter, so whichever line comes first wins. Ordering is not guaranteed, which is why it only shows up sometimes. It also means that while P2P DHCP is still running, we probe the home-network address instead of waiting for the P2P one.

Fix: only accept neighbours on the `p2p-*` group interface (excluding `p2p-dev-*`) in that MAC lookup.

Careful with the fallback right below it, `_get_peer_ip_from_p2p_iface`. That one exists for #44, where LG randomizes its MAC between discovery and connection so the MAC never matches. Adding the interface filter keeps that path intact: the MAC lookup finds nothing and the fallback still resolves the IP. Please verify that case stays working.

henriquejsza:
## Summary
- restrict MAC-based neighbour lookup to `p2p-*` group interfaces
- exclude `p2p-dev-*` and infrastructure interfaces from active probe targets
- preserve the interface-only fallback for TVs that randomize their MAC during group setup

The regression tests mock `ip neigh` output for a duplicate MAC across home/P2P interfaces, a `p2p-dev-*` entry, and the randomized-MAC fallback.

## Testing
- `PYSTRAY_BACKEND=dummy python -m unittest discover -s tests -v` (82 tests)
- `python -m compileall -q src tests`
- `git diff --check`

Fixes #85

### DIFF
--- a/src/wfd/p2p/addressing.py
+++ b/src/wfd/p2p/addressing.py
@@ -6,6 +6,10 @@ from typing import Optional
 from ..proc import _run
 
 
+def _is_p2p_group_iface(iface: str) -> bool:
+    return iface.startswith("p2p-") and not iface.startswith("p2p-dev-")
+
+
 def _get_peer_ip_from_arp(peer_mac: str) -> Optional[str]:
     """Return the IP for peer_mac from the kernel ARP/neighbour table."""
     if not shutil.which("ip"):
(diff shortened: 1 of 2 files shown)


IlyaP358 (approved):
lgtm! merging. 
Thanks especially for handling the randomized-MAC fallback from #44 and covering the whole chain through _wait_for_peer_ip rather than just the individual functions. That was the part I was most worried about breaking =]

PR #97, opened 2026-08-25
Validate wf-recorder before starting WFD capture

### LINKED ISSUE #96: AppImage wf-recorder wrapper is reported as available, then WFD falls back to X11 and produces a black screen
## Summary

The AppImage's `wf-recorder` wrapper is detected as available even when the system `wf-recorder` dependency is not installed. WFD negotiation then succeeds, but desktop capture falls back to X11 and the TV displays a black screen.

## Environment

- FluxCast v0.2.3 AppImage
- Arch Linux / Omarchy / Hyprland
- TCL TV using Wi-Fi Display (Miracast)

## Reproduction

1. Start a WFD cast from the AppImage on a Wayland/wlroots desktop without the system `wf-recorder` package.
2. Complete the TV connection.
3. Observe that RTSP M1/M3/M4/M5, SETUP, and PLAY all return `200 OK`.

The doctor reports `wf-recorder` as available because it finds the AppImage wrapper. At media start, the wrapper reports:

```text
[FluxCast] wf-recorder: not found on this system (install it and retry)
```

FluxCast then falls back to `x11grab`, which is black under Wayland. RTP packets continue to be sent, so the failure is easy to misdiagnose as a WFD negotiation problem.

Installing the system `wf-recorder` package resolves the issue.

## Expected behavior

The doctor should verify that `wf-recorder` can actually execute, and WFD should report the missing system dependency before starting a capture rather than launching a wrapper that exits immediately.

jainprashul:
## What changed

- Validate `wf-recorder --version` instead of treating an executable wrapper as sufficient.
- Use the same validation before starting the WFD wlroots capture pipeline.
- Add tests for a missing recorder, a failing AppImage-style wrapper, and a working recorder.
- Document the UFW rules needed for WFD RTSP and discovery:
  - `7236/tcp`
  - `5353/udp`

Fixes #96
Related to #98

## Verification

- `python -m unittest tests.test_wf_recorder tests.test_diagnostics`
- `python -m py_compile src/wfd/wf_recorder.py src/wfd/media/wlroots.py src/diagnostics.py`
- `git diff --check`

The complete local suite currently has one pre-existing environment failure because the optional `PIL` module is not installed for `tests/test_tray_config.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,11 +183,10 @@ Note: on KDE/GNOME Wayland, WFD auto backend now prefers `portal` first.
 
 Note: on **firewalld** systems, FluxCast opens the WFD RTSP port (`7236/tcp`) for the duration of a session and closes it on exit (no-op without firewalld; disable with `--wfd-no-firewall`). See [DOCUMENTATION.md](documentation/DOCUMENTATION.md) -> "WFD and firewalld".
 
-If you use **ufw**, open the WFD RTSP and discovery ports before connecting:
+If you use **ufw**, open the WFD RTSP port before connecting:
 
 ```bash
 sudo ufw allow 7236/tcp
-sudo ufw allow 5353/udp
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -180,6 +180,13 @@ Note: on KDE/GNOME Wayland, WFD auto backend now prefers `portal` first.
 
 Note: on **firewalld** systems, FluxCast opens the WFD RTSP port (`7236/tcp`) for the duration of a session and closes it on exit (no-op without firewalld; disable with `--wfd-no-firewall`). See [DOCUMENTATION.md](documentation/DOCUMENTATION.md) -> "WFD and firewalld".
 
+If you use **ufw**, open the WFD RTSP and discovery ports before connecting:
+
+```bash
+sudo ufw allow 7236/tcp
+sudo ufw allow 5353/udp
+```
+
 
 ## Documentation
 
--- a/documentation/DOCUMENTATION.md
+++ b/documentation/DOCUMENTATION.md
@@ -262,7 +262,6 @@ the port while a session is active and closes it on exit:
 
   ```bash
   sudo ufw allow 7236/tcp
-  sudo ufw allow 5353/udp
   ```
 
 Pass `--wfd-no-firewall` to skip this entirely and manage the firewall yourself.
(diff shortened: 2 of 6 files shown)


IlyaP358 (changes requested):
Hi. thanks, this is a real bug and the diagnosis is right: the wrapper is executable, so `shutil.which` reports success and the failure only surfaces later as a black screen on Wayland.

Three things before I can merge.

**Circular import.** The module-level `from wfd.wf_recorder import find_wf_recorder` in `diagnostics.py` collides with `wfd/__init__.py` importing `diagnostics`, and it breaks `--doctor` outright:

```
ImportError: cannot import name 'print_report' from 'diagnostics'
```

Your run does not catch it because `test_addressing.py` sorts first and loads `wfd`, so both modules are already in memory by the time `test_diagnostics` runs. Try `python3 -m unittest test_diagnostics` on its own. Moving the import into the two functions that use it is enough.

**`--version` is too strict.** I could not confirm that every wf-recorder release supports it, and if one does not, this disables the recorder for everyone on Wayland. My wrapper exits with exactly 127 when the system binary is missing, so rejecting only 127 targets it precisely and cannot misfire on a working binary.

**`5353/udp` is not needed.** That is mDNS, and nothing under `src/wfd/` uses it, no zeroconf or avahi anywhere. It only matters for Chromecast discovery through pychromecast. The ufw docs should mention `7236/tcp` only.

One small thing: please target `dev` rather than `main`, per CONTRIBUTING.

The tests themselves are good. And thanks for spotting the `test_tray_config` PIL failure, that one is mine from a refactor and I will fix it separately.

IlyaP358:
Hi @jainprashul,

I've switched the base branch of this PR to `dev` per our workflow guidelines in [CONTRIBUTING.md](https://github.com/IlyaP358/fluxcast/blob/main/CONTRIBUTING.md).

This caused a merge conflict in `README.md` because I recently updated the supported distros list there. Whenever you have some time, could you please rebase your branch on `dev`, resolve the conflict (keeping the new distros), and take a look at the `--doctor` feedback?

Take your time, there is no rush at all. Thanks!

jainprashul:
@IlyaP358 , Please review again.

IlyaP358 (approved):
lgtm👍Thanks for the correction and for your help, I'll definitely add you to the [contributors page](https://fluxcast.dev/contributors.html) =] , merging

PR #101, opened 2026-08-26
Add German translation

therealresonix:
This adds a German translation.

It's quite possible that the translation isn't 100 percent accurate, since the English version isn't entirely clear and could mean two different things in some places.


IlyaP358 (approved):
lgtm! merging.
Thanks a lot for this. I will definitely add you to the [contributors page](https://fluxcast.dev/contributors.html), you deserve it =]

Issue #102, opened 2026-08-26
DLNA/Cast stream server listens on every interface and serves any client


IlyaP358:
In DLNA and Cast modes the desktop stream is served to anyone who asks. While a session is running, any machine on the network can open `vlc http://<your-ip>:8080/` and watch. There is nothing to guess, the port defaults to 8080 and the paths are fixed.

Three small things add up to it. `main.py:355` binds the server to `0.0.0.0`, so it listens on every interface including VPN and the transient P2P one, even though the address the TV is actually told to use is already computed at `main.py:287`. The URL we hand the device carries a session prefix (`main.py:288`), but `session_id` never appears in `server.py`, so `/live.ts` and a bare `/` are served without it and the prefix does nothing. And `client_address` is only written to the log (`server.py:350`), never compared against the target device.

The same applies locally: `HLS_DIR` is a fixed `/tmp/fluxcast` created with default permissions, so other users on the machine can read the segments off disk.

Fixing it means binding to the address from `main.py:287`, actually enforcing the session prefix and generating it with `secrets.token_urlsafe` instead of `int(time.time())`, refusing requests that don't come from the target device, and creating the segment directory per session with mode `0700`. None of that changes anything for the user, the TV already receives that exact address and path from us.

WFD is unaffected, it never reaches this server.

To check: with a session running, `curl -sI http://<your-ip>:8080/` from another machine should be refused while the TV keeps playing.

PR #104, opened 2026-08-26
Fix wpa_supplicant D-Bus policy over-grant 

### LINKED ISSUE #103: wpa_supplicant D-Bus policy grants Properties.Set to every local account
`meta/zz-dev.fluxcast.wpa-supplicant.conf` uses `<policy context="default">`, which is every account on the machine. wpa_supplicant denies that context on purpose, and our file sorts last and overrides it.

Two calls as a normal user show it. One the policy allows:

```
$ gdbus call --system --dest fi.w1.wpa_supplicant1 --object-path /fi/w1/wpa_supplicant1 \
    --method org.freedesktop.DBus.Properties.Set fi.w1.wpa_supplicant1 NoSuchProperty "<true>"
Error: GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: No such property
```

One it doesn't:

```
$ gdbus call --system --dest fi.w1.wpa_supplicant1 --object-path /fi/w1/wpa_supplicant1 \
    --method fi.w1.wpa_supplicant1.GetInterface "nosuchiface0"
Error: GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: Sender is not authorized to send message
```

The second was dropped by the bus, the first reached the daemon and got answered. So any unprivileged process can write wpa_supplicant properties, `DebugShowKeys` for instance, and FluxCast doesn't even need to be running. Service accounts included, so an exploit in some daemon running under its own user gets a path to the root supplicant.

Fluxcast write one property, `P2PDeviceConfig`, for the device name and for the GO intent LG needs (`src/wfd/p2p/device.py:64` and `:113`). D-Bus can't limit a grant to a single property, only to who may send it, so the fix is scoping the policy to an admin group instead of `default`: `wheel` on Arch and Fedora, `sudo` on Debian and Ubuntu. Those users can already sudo, so it gives them nothing new, and a desktop user is already a member.

Both call sites already fall back to a warning, but losing the GO intent can quietly break LG sinks, so `--doctor` should say whether you're in the group. The file ships from `meta/install.sh`, `meta/pypi/pypi_sysinstall.py`, `appimage/fluxcast-setup.sh` and the AUR package.

jainprashul:
## Summary
- Scope the FluxCast wpa_supplicant D-Bus policy to `wheel`/`sudo` instead of `context="default"`, so unprivileged local accounts can no longer call `Properties.Set` on the system supplicant ([#103](https://github.com/IlyaP358/fluxcast/issues/103)).
- Add a `--doctor` check that reports whether the current user is in an admin group (needed for device name / GO intent).
- Refresh the AppImage first-run setup when the installed policy content changes, so existing insecure installs get upgraded.

## Test plan
- [x] `python3 -m unittest tests.test_diagnostics -v`
- [x] Install/update `meta/zz-dev.fluxcast.wpa-supplicant.conf` and `sudo systemctl reload dbus`
- [x] As a normal non-admin user, confirm `Properties.Set` is `AccessDenied` (bus drop), while `wheel`/`sudo` members can still set `P2PDeviceConfig`
- [x] `python3 src/main.py --doctor` shows `wpa D-Bus group` ok/warn based on membership
- [x] AppImage/setup path replaces an older `context="default"` policy file when content differs

Fixes #103

### DIFF
--- a/appimage/fluxcast-setup.sh
+++ b/appimage/fluxcast-setup.sh
@@ -6,11 +6,18 @@ DBUS_CONF_NAME="zz-dev.fluxcast.wpa-supplicant.conf"
 DBUS_CONF_DEST="/usr/share/dbus-1/system.d/${DBUS_CONF_NAME}"
 DBUS_CONF_SRC="${APPDIR}/meta/${DBUS_CONF_NAME}"
 
-[ -f "${DBUS_CONF_DEST}" ] && exit 0   # already installed, nothing to do
 [ -f "${DBUS_CONF_SRC}" ]  || exit 0   # config missing from AppDir, skip
+# Refresh when missing or content changed (e.g. insecure default-context grant).
+if [ -f "${DBUS_CONF_DEST}" ] && cmp -s "${DBUS_CONF_SRC}" "${DBUS_CONF_DEST}"; then
+    exit 0
+fi
 
 echo ""
-echo "[FluxCast] First-time setup: installing DBus policy for Wi-Fi Direct..."
+if [ -f "${DBUS_CONF_DEST}" ]; then
+    echo "[FluxCast] Updating DBus policy for Wi-Fi Direct..."
+else
+    echo "[FluxCast] First-time setup: installing DBus policy for Wi-Fi Direct..."
+fi
 
 # FUSE mounts are not accessible by root. Copy to /tmp first so sudo can read it.
 TMP_CONF="$(mktemp /tmp/fluxcast-dbus-XXXXXX.conf)"
--- a/meta/zz-dev.fluxcast.wpa-supplicant.conf
+++ b/meta/zz-dev.fluxcast.wpa-supplicant.conf
@@ -2,7 +2,22 @@
  "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
  "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
 <busconfig>
-  <policy context="default">
+  <!--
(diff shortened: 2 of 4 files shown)


IlyaP358 (changes requested):
The D-Bus part is correct. I verified it: the two-group policy loads, dbus-daemon logs `Unknown group "sudo"` on Arch where that group doesn't exist and starts normally, and `--doctor` reports the new row.

But this PR is titled "issue #103", says `Fixes #103`, and then quietly does #102 as well in the second commit, and you submitted that same commit again as #105. Please don't send the same change in two PRs. Drop the second commit here so this is #103 only.

And a general request: please don't vibe-code changes into this project. Before you open a PR, actually run the thing and confirm the core paths still work. `python3 src/main.py --protocol dlna` takes three minutes. A green unit-test suite is not evidence that FluxCast still functions, as #105 shows.

One suggestion on the doctor check while you're in there: it infers access from group membership, so it reports ok even when the policy file was never installed or dbus was never reloaded. One call makes it authoritative: `Properties.Set` on a property that doesn't exist returns `AccessDenied` when the bus blocks it, and `InvalidArgs: No such property` when the grant is live.

jainprashul:
Dropped the second commit so this PR is #103 only. The HLS/stream-server access work stays in #105.

Also updated `--doctor` as suggested: it now probes `Properties.Set` on a non-existent property (`AccessDenied` vs `InvalidArgs`) instead of inferring access from group membership, so a missing policy file or an unreloaded dbus is reported correctly.

Ready for another look.

IlyaP358 (approved):
Looks great 👍 
Merging this into dev

PR #105, opened 2026-08-26
Restrict DLNA/Cast stream server access (issue #102)

### LINKED ISSUE #102: DLNA/Cast stream server listens on every interface and serves any client

jainprashul:
## Summary
- Bind the DLNA/Cast HTTP stream server to the advertised LAN address instead of `0.0.0.0`, and require an unpredictable `secrets.token_urlsafe` session URL prefix.
- Allow only the selected TV/Chromecast client IP (403 until set / for other hosts), and store HLS segments under a per-session directory with mode `0700`.
- Adds unit tests covering bind address, session prefix, client ACL, and directory permissions. Fixes #102.

## Test plan
- [ ] `python -m unittest tests.test_stream_access -v`
- [ ] Start a DLNA or Cast session; confirm the TV still plays from the session URL
- [ ] From another machine on the LAN: `curl -sI http://<host>:8080/` is refused (404/403) while the TV keeps playing
- [ ] Confirm `/tmp/fluxcast/<session-id>/` exists with mode `0700`
- [ ] Confirm WFD path is unchanged

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/capture/capture.py
+++ b/src/capture/capture.py
@@ -7,7 +7,19 @@ import sys
 import time
 from typing import Any, Optional, NamedTuple
 
+import server as stream_server
+
 FFMPEG_PATH = shutil.which("ffmpeg") or "/usr/sbin/ffmpeg"
+
+
+def _prepare_capture_hls_dir() -> str:
+    hls_dir = stream_server.HLS_DIR
+    os.makedirs(hls_dir, mode=0o700, exist_ok=True)
+    try:
+        os.chmod(hls_dir, 0o700)
+    except OSError:
+        pass
+    return hls_dir
 
 
 class Monitor(NamedTuple):
@@ -272,6 +284,7 @@ def _start_capture_wf_recorder(
         out_w, out_h = 1920, 1080
 
     audio_monitor = _default_audio_monitor()
+    hls_dir = _prepare_capture_hls_dir()
 
     cmd = [
         "wf-recorder",
@@ -292,17 +305,12 @@ def _start_capture_wf_recorder(
         "-p", "hls_flags=append_list",
         "-p", "pix_fmt=yuv420p",
         "-p", "profile=main",
-        "-f", "/tmp/fluxcast/stream.m3u8",
+        "-f", os.path.join(hls_dir, "stream.m3u8"),
         "-o", monitor.name,
     ]
 
     print(f"[FluxCast] Capturing Wayland monitor : {monitor.name} ({src_res})")
     print(f"[FluxCast] Output: {out_w}x{out_h}, audio: {audio_monitor}")
-
-    hls_dir = "/tmp/fluxcast"
-    if os.path.exists(hls_dir):
-        shutil.rmtree(hls_dir)
-    os.makedirs(hls_dir, exist_ok=True)
 
     try:
         process = subprocess.Popen(
@@ -376,10 +384,7 @@ def _start_capture_portal(
         except (ValueError, AttributeError):
             pass
 
-    hls_dir = "/tmp/fluxcast"
-    if os.path.exists(hls_dir):
-        shutil.rmtree(hls_dir)
-    os.makedirs(hls_dir, exist_ok=True)
+    hls_dir = _prepare_capture_hls_dir()
 
     audio_monitor = _default_audio_monitor()
     audio_enc = (
@@ -503,11 +508,7 @@ def _start_capture_x11grab(
     src_res = f"{monitor.width}x{monitor.height}"
     out_res = output_resolution or src_res
     audio_monitor = _default_audio_monitor()
-    hls_dir = "/tmp/fluxcast"
-
-    if os.path.exists(hls_dir):
-        shutil.rmtree(hls_dir)
-    os.makedirs(hls_dir, exist_ok=True)
+    hls_dir = _prepare_capture_hls_dir()
 
     cmd = [
         ffmpeg,
--- a/src/main.py
+++ b/src/main.py
@@ -65,7 +65,15 @@ import sys
 import termios
 
 from capture import prompt_monitor, start_capture, stop_capture
-from server import HLS_DIR, CorsHLSRequestHandler, HLSRequestHandler, StreamServer
+from server import (
+    HLS_DIR,
+    CorsHLSRequestHandler,
+    HLSRequestHandler,
+    StreamServer,
+    device_client_ip,
+    new_session_id,
+    prepare_hls_dir,
+)
 
 
 def get_local_ip() -> str:
@@ -285,7 +293,8 @@ def main() -> None:
         return
 
     host = args.host or get_local_ip()
-    session_id = f"session-{int(time.time())}"
+    session_id = new_session_id()
+    prepare_hls_dir(session_id)
     if args.transport == "live-ts":
         stream_name = "live.ts"
     elif args.transport == "progressive-ts":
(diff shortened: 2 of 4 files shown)


IlyaP358 (changes requested):
This breaks DLNA and Cast completely.

`main.py` imports `HLS_DIR` by value, so `prepare_hls_dir()` reassigning `server.HLS_DIR` never reaches it. `_wait_for_hls_segments` looks in `/tmp/fluxcast` while capture writes to `/tmp/fluxcast/<session-id>/`. It times out after 30 seconds and calls `shutdown()`, so every run with `--protocol dlna` or `--protocol cast` dies with "ffmpeg produced no playable HLS segments". Read `server.HLS_DIR` dynamically, the way `capture.py` already does.

Your Test plan says "Start a DLNA or Cast session; confirm the TV still plays from the session URL". That step cannot have passed, so it was not run. Please don't submit a test plan describing checks you didn't do, and please don't vibe-code security changes into the two protocols users actually rely on without confirming they still work end to end. The unit tests here are all green and the feature is dead, because nothing in the suite imports `main`. Add a test that does.

Two smaller ones: the `elif basename in ("live.ts", "progressive.ts")` branch in `_local_path` is unreachable, `do_GET` routes those before `_serve_file` runs. And `os.makedirs(..., mode=0o700)` applies the mode only to the leaf directory, so `/tmp/fluxcast` itself stays 0755.

The rest of the approach is right. I checked that ffmpeg writes relative segment names into the playlist in both capture paths, so the session prefix resolves correctly, and binding to the advertised address instead of `0.0.0.0` is the correct fix.

jainprashul:
> This breaks DLNA and Cast completely.
> 
> `main.py` imports `HLS_DIR` by value, so `prepare_hls_dir()` reassigning `server.HLS_DIR` never reaches it. `_wait_for_hls_segments` looks in `/tmp/fluxcast` while capture writes to `/tmp/fluxcast/<session-id>/`. It times out after 30 seconds and calls `shutdown()`, so every run with `--protocol dlna` or `--protocol cast` dies with "ffmpeg produced no playable HLS segments". Read `server.HLS_DIR` dynamically, the way `capture.py` already does.
> 
> Your Test plan says "Start a DLNA or Cast session; confirm the TV still plays from the session URL". That step cannot have passed, so it was not run. Please don't submit a test plan describing checks you didn't do, and please don't vibe-code security changes into the two protocols users actually rely on without confirming they still work end to end. The unit tests here are all green and the feature is dead, because nothing in the suite imports `main`. Add a test that does.
> 
> Two smaller ones: the `elif basename in ("live.ts", "progressive.ts")` branch in `_local_path` is unreachable, `do_GET` routes those before `_serve_file` runs. And `os.makedirs(..., mode=0o700)` applies the mode only to the leaf directory, so `/tmp/fluxcast` itself stays 0755.
> 
> The rest of the approach is right. I checked that ffmpeg writes relative segment names into the playlist in both capture paths, so the session prefix resolves correctly, and binding to the advertised address instead of `0.0.0.0` is the correct fix.


Fixed in 4b7dc98:
- `_wait_for_hls_segments` now reads `server.HLS_DIR` dynamically (same pattern as `capture.py`)
- Added `MainHlsDirBindingTest` that imports `main` and asserts segment wait follows `prepare_hls_dir`
- Removed the unreachable `_local_path` branch
- `prepare_hls_dir` now `chmod`s both `HLS_BASE` and the session dir to `0700`
`python -m unittest tests.test_stream_access -v` passes.

jainprashul:
python -m unittest tests.test_stream_access -v
test_cast_device_ip (tests.test_stream_access.DeviceIpTest.test_cast_device_ip) ... ok
test_dlna_device_ip_from_location (tests.test_stream_access.DeviceIpTest.test_dlna_device_ip_from_location) ... ok
test_wait_for_hls_segments_follows_prepare_hls_dir (tests.test_stream_access.MainHlsDirBindingTest.test_wait_for_hls_segments_follows_prepare_hls_dir)
main must read server.HLS_DIR dynamically, not import it by value. ... ok
test_new_session_id_uses_secrets (tests.test_stream_access.SessionSetupTest.test_new_session_id_uses_secrets) ... ok
test_prepare_hls_dir_uses_session_subdir_mode_0700 (tests.test_stream_access.SessionSetupTest.test_prepare_hls_dir_uses_session_subdir_mode_0700) ... ok
test_allows_correct_session_and_client (tests.test_stream_access.StreamAccessTest.test_allows_correct_session_and_client) ... ok
test_binds_only_to_requested_host (tests.test_stream_access.StreamAccessTest.test_binds_only_to_requested_host) ... ok
test_rejects_bare_root_without_session_prefix (tests.test_stream_access.StreamAccessTest.test_rejects_bare_root_without_session_prefix) ... ok
test_rejects_before_allowed_client_is_set (tests.test_stream_access.StreamAccessTest.test_rejects_before_allowed_client_is_set) ... ok
test_rejects_disallowed_client (tests.test_stream_access.StreamAccessTest.test_rejects_disallowed_client) ... ok
test_rejects_unprefixed_live_ts (tests.test_stream_access.StreamAccessTest.test_rejects_unprefixed_live_ts) ... ok
test_rejects_wrong_session_prefix (tests.test_stream_access.StreamAccessTest.test_rejects_wrong_session_prefix) ... ok

----------------------------------------------------------------------
Ran 12 tests in 4.093s

OK
 ready! (2 segments, 0.0s)
[FluxCast Server] 404 127.0.0.1 -> /
[FluxCast Server] 403 127.0.0.1 -> /session-testtoken/stream.m3u8
[FluxCast Server] 403 127.0.0.1 -> /session-testtoken/stream.m3u8
[FluxCast Server] 404 127.0.0.1 -> /live.ts
[FluxCast Server] 404 127.0.0.1 -> /session-other/stream.m3u8

IlyaP358 (changes requested):
Hi, the access control checks out. I verified the client ACL, the session prefix, and path traversal against the running server, all behave correctly.

There is one regression though, in the `--host` flag.

1. It was documented as the IP to advertise in the stream URL, and the server always bound `0.0.0.0`. Now the same value is used for the bind, so any address that doesn't exist on the machine fails: an external or NAT IP gives `OSError: [Errno 99] Cannot assign requested address`, and a hostname gives `gaierror`. Both are unhandled.
2. `stream_server.start()` on `main.py:374` isn't wrapped in a try, and `start_capture()` has already run by then, so the crash leaves an orphaned screen capture process behind and the terminal in a broken state, because `shutdown()` only runs on SIGINT/SIGTERM.

Please wrap the server start so a bind error goes through `shutdown()` with a readable message instead of a traceback, and update the flag's help string, since it now controls the bind address too and not just what gets advertised.

Everything else looks solid. Once that's in, I'm ready to merge.

Issue #72, opened 2026-07-16
LG webOS TV discovered and P2P connects, but no RTSP session starts


Bibergoahead:
Bug Report: LG webOS TV discovered and P2P connects, but no RTSP session starts

Environment
- FluxCast: fluxcast-git 0.1.5.r0.g38fdbc2-1
- Distribution: EndeavourOS (Arch Linux)
- Desktop: KDE Plasma (X11)
- Kernel: 6.18.38-3-lts
- Python: 3.14.6
- NetworkManager: 1.56.1
- Wi-Fi adapter: Intel AX201
- TV: LG OLED55B87LC (webOS)

Doctor output
All required checks pass:
- ffmpeg ✔
- NetworkManager ✔
- dnsmasq ✔
- iw ✔
- wpa_supplicant P2P ✔
- WFD capability ✔
- x11grab ✔
- H.264 encoders ✔
- Audio capture ✔

Only warnings:
- wf-recorder not installed (expected, using X11)
- dbus_next missing
- firewalld warning

Device discovery
Running:
fluxcast --wfd-scan
finds the TV every time:

[1] 00:51:ED:35:0F:AE  lgoled55b87lc
    WFD capability data detected

Connection attempt
Running:
fluxcast --protocol wfd --wfd-peer 1
results in:
- NetworkManager activates P2P successfully.
- P2P link becomes active.
- RTSP server starts on port 7236.
- firewalld opens port 7236.
- FluxCast waits for the TV RTSP session.

Then:
Active probe: TV=10.42.0.182
TV port 7236 refused — Sink-only device; waiting for its passive connection to us.

The TV never opens an RTSP connection.

Network verification
P2P interface:
10.42.0.1/24

TV:
10.42.0.182

Ping works:
0% packet loss

RTSP server is listening:
ss -tln | grep 7236
shows:
LISTEN 0 5 0.0.0.0:7236

tcpdump:
sudo tcpdump -i p2p-wlan0-1 host 10.42.0.182
shows no packets from the TV attempting an RTSP connection.

Additional observations
Using:
fluxcast --protocol wfd --wfd-peer 00:51:ED:35:0F:AE
returns:
ERROR: No peer matched selector: 00:51:ED:35:0F:AE
although the previous --wfd-scan clearly lists the same MAC address.

Similarly,
--wfd-peer 1
sometimes fails with:
Peer index out of range: 1
immediately after a successful scan showing index 1.

[...]

PR #107, opened 2026-08-29
Add a direct wpa_supplicant P2P backend for WFD (--wfd-p2p-backend wpas)

Jemo121:
## Summary

- New opt-in P2P backend, `--wfd-p2p-backend wpas`, that talks to wpa_supplicant's own D-Bus interface directly instead of going through NetworkManager. The default (`nm`) is untouched - this is purely additive.
- It gives control that NetworkManager's own P2P API doesn't expose: the WFD Device Info IE content, GO-intent timing, and the P2P operating channel.
- New `--wfd-p2p-channel {1,6,11}` flag (only meaningful with `wpas`) forces the P2P group onto a 2.4GHz channel. Some WFD sinks only support Wi-Fi Direct on 2.4GHz - if the driver picks 5GHz on its own, GO Negotiation completes fine, but the sink never actually shows up at the 802.11 level.
- Small widening of `meta/zz-dev.fluxcast.wpa-supplicant.conf`: the four P2PDevice actions the new backend needs (`Find`/`Connect`/`StopFind`/`GroupRemove`) are now also granted to the `netdev` group, not just root - the same group wpa_supplicant's own control socket is already scoped to. `_gdbus_call()` tries unprivileged first and only falls back to `sudo` on an actual `AccessDenied`, so this doesn't change behavior for anyone already running as root.

## Why

I ran into a Miracast/WFD sink where the NetworkManager P2P path kept failing to configure IP addressing correctly, and NetworkManager's own API gave me no visibility into what was actually happening during negotiation. Talking to wpa_supplicant directly let me watch (and control) each step, which surfaced two real, fixable issues along the way: the WFD IE wasn't declaring the correct device role on this path, and this particular Wi-Fi chipset only forms a usable P2P group on 2.4GHz with some sinks. Both are fixed here, and the new backend should be generally useful for anyone debugging a stubborn P2P connection, not just my specific case.

## What this doesn't fix

The `netdev` policy change only covers P2P negotiation. `wpas_ip.py`'s IP-layer work (`ip`, `dnsmasq`, `dhcpcd`) is raw shell commands, not D-Bus, so it still needs real root regardless of group membership - a `netdev`-only user without sudo can discover and negotiate but can't complete a connection end-to-end today. Closing that properly would mean a small privileged helper (a setuid binary, or a tiny root-owned daemon this process talks to over a socket). I kept that out of this PR to keep it focused, but happy to take a swing at it as a follow-up if that's something you'd want.

## Testing

Tested against a real WFD sink over several sessions - GO Negotiation, WPS provisioning, and 802.11 association all complete correctly through this backend, and the channel-forcing behavior does what it says (checked via `iw station dump` and wpa_supplicant's own debug log). The specific sink I tested against turned out to have its own unrelated compatibility issue after association - a client-side WPS quirk, not anything on FluxCast's side - so I never got a full end-to-end cast against that particular unit. I'll file that separately against wpa_supplicant/hostap upstream, since it's not something this project can fix; mentioning it here mostly in case anyone else runs into the same wall.

Thanks for maintaining this project - happy to adjust anything here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,9 @@ python3 src/main.py --protocol wfd --wfd-p2p-backend wpas --wfd-p2p-channel 6
 ```
 
 Needs the D-Bus policy in `meta/zz-dev.fluxcast.wpa-supplicant.conf`
-installed to `/usr/share/dbus-1/system.d/` (root, or a user in the `netdev`
-group, can then run it without sudo - see the file for details).
+installed to `/usr/share/dbus-1/system.d/`. On Debian/Ubuntu, a user in the
+`netdev` group can then run it without sudo; elsewhere (including Arch)
+it falls back to sudo - see the file for details.
 
 ## What Works Best
 
--- a/README.md
+++ b/README.md
@@ -84,6 +84,19 @@ python3 src/main.py --protocol wfd --wfd-capture-backend wf-recorder
 python3 src/main.py --protocol wfd --wfd-capture-backend x11grab
 ```
 
+WFD P2P connection troubleshooting - talk to wpa_supplicant directly instead
+of going through NetworkManager, and optionally pin the group to a 2.4GHz
+channel for sinks that don't support Wi-Fi Direct on 5GHz:
+
+```bash
+python3 src/main.py --protocol wfd --wfd-p2p-backend wpas
+python3 src/main.py --protocol wfd --wfd-p2p-backend wpas --wfd-p2p-channel 6
+```
+
+Needs the D-Bus policy in `meta/zz-dev.fluxcast.wpa-supplicant.conf`
+installed to `/usr/share/dbus-1/system.d/` (root, or a user in the `netdev`
+group, can then run it without sudo - see the file for details).
+
 ## What Works Best
 
 ### WFD (Primary)
--- a/documentation/DOCUMENTATION.md
+++ b/documentation/DOCUMENTATION.md
@@ -69,6 +69,8 @@ python3 src/main.py --protocol cast
 - `--wfd-interface IFACE`
 - `--wfd-timeout SEC`
 - `--wfd-go-intent 0-15`
+- `--wfd-p2p-backend nm|wpas` talk to wpa_supplicant directly instead of NetworkManager (default `nm`)
+- `--wfd-p2p-channel 1|6|11` force the P2P group onto this 2.4GHz channel (`wpas` backend only)
 - `--wfd-uibc` enable the input back channel (control the desktop from the sink)
 - `--wfd-monitor NAME` **deprecated** alias for `--monitor` (kept for compatibility)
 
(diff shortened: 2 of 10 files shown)


IlyaP358 (changes requested):
Hi, thanks, this is a solid PR and the write-up made review easy. The explicit role handling at the IP layer is the part I really want, along with the visibility it gives when a P2P link misbehaves. Three things before I merge.

**dnsmasq isn't scoped the way you intended.** `--dhcp-range=set:wfdsink,...` labels whoever takes an address from the range, it doesn't limit the range to tagged clients. Matching needs `tag:`. Combined with `--interface={physical_iface}`, a GO-role session answers DHCP on the user's home network too. Anything that asks can get an address in 192.168.49.0/24 with our laptop as its default route and DNS server, and since `--port=0` disables DNS we answer neither. That device loses its network until the lease expires. It is a race against the real router, so it surfaces as an intermittent "my phone drops off wifi when I cast", which is horrible to diagnose. Your comment describes exactly the behaviour you were trying to prevent, so I think it is just the one word: `--dhcp-range=tag:wfdsink,...`.

**netdev.** You are right on Debian and Ubuntu, but Arch is this project's main platform and there it does not hold: no `netdev` group exists, none of the four wpa_supplicant units has a `Group=` line, and `/run/wpa_supplicant` is `drwxr-x--- root root`, so no group has that access to begin with. The policy block itself is harmless, dbus just logs `Unknown group` and carries on. But the comment saying it extends an existing trust boundary, and the README line about running without sudo, are both Debian specific. Please reword them to say where that applies and that everyone else falls back to sudo.

**`--wfd-p2p-channel`.** You checked the forcing works and I believe you. But `--wfd-go-intent` defaults to 0 so the sink becomes GO, and per your own docstring the channel applies when we are GO. Did your runs end up with you as GO? If so, users on the default will not see the flag do anything, and someone with a 2.4GHz-only sink will conclude it is broken. A warning when a channel is set alongside a low GO intent, plus a note about the `--wfd-go-intent 15` combination, would cover it. Just do not wire the two together automatically, `go_intent 0` is the LG fix from #72 and raising it silently would break those sinks.

Smaller: `TimeoutExpired` from the privileged `Find` and `Connect` calls is not caught, so a timeout comes out as a traceback instead of a `WFDNotReady`. And there are no tests for the new modules, something around the WFD IE bytes and the channel argument building would do.

Yes to the privileged-helper follow-up you offered, but let's land this first.

Jemo121:
Thanks for the thorough review - genuinely useful catches, especially the dnsmasq one.

**dnsmasq scoping.** You're right, and it's a real bug, not just a wording nit - fixed to `tag:wfdsink`. Added a regression test for it too, since it's exactly the kind of thing that fails silently.

**netdev.** Fair, I was thinking Debian/Ubuntu and wrote it as if that applied everywhere. Reworded the policy file's comment, the `_gdbus_call` docstring, and both docs to say this is a Debian/Ubuntu convention specifically, and that everyone else (Arch included) falls back to sudo - same as before this policy existed.

**`--wfd-p2p-channel`.** You called it exactly right - my runs happened to land on GO regardless of intent (specific to the sink I was testing against), so I didn't hit this. Added a warning once the actual post-negotiation role is known (checking the real role rather than guessing from go_intent beforehand, since intent alone doesn't reliably predict who ends up as GO), plus a doc note about pairing with `--wfd-go-intent 15` for deterministic control. Didn't wire them together automatically, per your note on #72.

**Timeout handling + tests.** Both done - `_gdbus_call` now converts `TimeoutExpired` to `WFDNotReady` for both the privileged and unprivileged paths, and I added `tests/test_wpas.py` covering the WFD IE bytes, the channel D-Bus argument, the timeout conversion, and the dnsmasq tag regression specifically.

All pushed. Let me know if anything needs another pass.

IlyaP358 (approved):
lgtm👍

Thanks a lot for this one. You took every point seriously, and basing the channel warning on the real post-negotiation role instead of the intent is better than what I asked for. Merging now!

You'll be added to [https://fluxcast.dev/contributors.html](https://fluxcast.dev/contributors.html), you've genuinely earned it! =]

PR #108, opened 2026-09-02
Fix wpas backend silently breaking under #104's wheel/sudo D-Bus policy

Jemo121:
## Summary

`dev` currently has a working combination that produces a silent regression: #104 correctly restricted wpa_supplicant's `Properties.Get`/`Set` to `wheel`/`sudo`, and #107 (mine, merged just before this) added a `netdev`-scoped grant for the four P2P action methods. Neither PR was written against the other's final state, so the interaction fell through: every `Properties.Get`/`Set` call the `wpas` backend makes was still unprivileged, which now fails for a `netdev`-only caller with no `wheel`/`sudo` - exactly the persona #107's own policy block was written to support.

## What actually breaks

- `_p2p_device_iface_paths`, `_wpas_get_property`, `_list_wpas_interfaces` all treat a denied `Get` as "nothing found" and return empty - so peer discovery just looks like it silently times out, no error at all.
- `_read_p2p_go_intent` returning `None` on denial means a GO intent change never gets restored on cleanup, quietly left changed.
- `_set_wfd_ies` is the loud one: it raises on failure, so it aborts every single connection attempt for that caller outright.

## Fix

Every wpa_supplicant `Properties.Get`/`Set` call in `wpas.py` and `device.py` now passes `privileged=True`, so `_gdbus_call`'s existing sudo-retry-on-`AccessDenied` path (already proven for `Find`/`Connect`/`StopFind`/`GroupRemove` in #107) covers these too. Added a short note to the policy file's comment explaining the interaction for whoever reads it next.

## Testing

Added two regression tests rather than trusting the mechanism transfers by inspection: `_set_wfd_ies` (the hard failure) and `_p2p_device_iface_paths` (the lookup nearly everything else in the backend depends on) both assert the sudo retry actually fires on a simulated `AccessDenied`. Full suite still green (93 tests, one pre-existing unrelated failure from a missing `pystray` import in this sandbox).

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,8 +95,9 @@ python3 src/main.py --protocol wfd --wfd-p2p-backend wpas --wfd-p2p-channel 6
 
 Needs the D-Bus policy in `meta/zz-dev.fluxcast.wpa-supplicant.conf`
 installed to `/usr/share/dbus-1/system.d/`. On Debian/Ubuntu, a user in the
-`netdev` group can then run it without sudo; elsewhere (including Arch)
-it falls back to sudo - see the file for details.
+`netdev` group can run it without sudo, courtesy of the `netdev` grant in
+the distro's own `wpa_supplicant.conf`; elsewhere (including Arch) it falls
+back to sudo - see the file for details.
 
 ## What Works Best
 
--- a/documentation/DOCUMENTATION.md
+++ b/documentation/DOCUMENTATION.md
@@ -205,9 +205,10 @@ and protocol selection remain controlled by the tray and cannot be set here.
     `--wfd-p2p-channel` below. Also useful when a P2P connection needs
     closer debugging, since it logs each step of the raw negotiation.
   - Needs the D-Bus policy in `meta/zz-dev.fluxcast.wpa-supplicant.conf`
-    installed. On Debian/Ubuntu, a user in the `netdev` group can then run
-    it without sudo; elsewhere (including Arch, this project's primary
-    platform) it falls back to sudo - see the file for details.
+    installed. On Debian/Ubuntu, a user in the `netdev` group can run it
+    without sudo, courtesy of the `netdev` grant in the distro's own
+    `wpa_supplicant.conf`; elsewhere (including Arch, this project's
+    primary platform) it falls back to sudo - see the file for details.
 - `--wfd-p2p-channel`
   - Forces the P2P group onto channel `1`, `6`, or `11` (2.4GHz) instead of
     letting the driver pick. Only used by `--wfd-p2p-backend wpas`.
--- a/meta/zz-dev.fluxcast.wpa-supplicant.conf
+++ b/meta/zz-dev.fluxcast.wpa-supplicant.conf
@@ -46,7 +46,11 @@
 
     This covers P2P negotiation only. The IP-layer side of the wpas backend
     (raw ip/dnsmasq/dhcpcd commands, not D-Bus calls) still needs real root
-    regardless of group membership.
+    regardless of group membership. It also doesn't cover Properties.Get/Set
+    (wheel/sudo only, above) - a netdev-only caller with neither still needs
+    the sudo fallback in _gdbus_call for those, which is why every wpas.py
+    and device.py call site that touches a Properties.Get/Set on this
+    service passes privileged=True.
   -->
   <policy group="netdev">
     <allow send_destination="fi.w1.wpa_supplicant1"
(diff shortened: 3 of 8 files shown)


IlyaP358 (changes requested):
Hi. thanks for pr and good catch on the interaction. The diagnosis is right, the netdev block never covered `Properties.Get`/`Set`, so a netdev-only caller does get denied, and the three failure modes you list are exactly what the code does.

One problem with the fix though. `_set_p2p_device_name` is called at `session.py:73`, before the backend branch at `:138`, so it runs for every WFD session regardless of backend, and `_set_p2p_go_intent` runs on the `nm` branch at `:151`. With `privileged=True` on the shared helpers, anyone outside `wheel`/`sudo` now gets a sudo password prompt on the default NetworkManager path where they previously just got a cosmetic warning. I measured it with a mocked denial: two sudo invocations per session at 3 seconds each, so about six seconds of stalling, and then the same two warnings as before. #113 is a user in exactly that position.

Could you add a `privileged` parameter to the shared `device.py` helpers, defaulting to `False`, and pass `True` only from `wpas.py`? `_set_p2p_oper_channel` can stay unconditional since only `wpas.py` calls it. That leaves the `nm` path untouched and still fixes the wpas one.

Worth noting for scope: a netdev-only caller can't finish a session anyway, since `wpas_ip.py` shells out to `sudo` for `ip`/`dnsmasq`/`dhcpcd` in a dozen places, as you noted yourself in #107. So this is really about failing clearly rather than enabling that persona, which is fine, it just doesn't justify touching the default path.

Tests look good, and I like that both new ones assert the retry actually fires instead of trusting it by inspection.

IlyaP358:
One more thing I found after posting the review, worth folding in while you're in there.

Debian's own `wpa_supplicant.conf` has a `<policy group="netdev">` with a bare `send_destination` and no member restriction, so netdev members already get unrestricted access to the service there. Our policy file only contains `allow` rules, so it can't narrow that. Which means the netdev block we added in #107 is redundant on Debian/Ubuntu and inert on Arch, where the group doesn't exist at all.

So the persona in your summary is already covered by the distro on Debian, and on Arch it comes down to a user outside `wheel` who has sudo rights configured individually. Still worth fixing for that case, since the wpas IP layer assumes sudo anyway, just much narrower than it looked.

If you agree, drop the netdev block from the policy in this PR as well and replace its comment with a short note on why it isn't needed. That also gets rid of the "unknown group" line dbus logs on Arch.

Jemo121:
Both done, thanks - the `session.py:73` point was the part I'd missed.

`privileged` is now a parameter on the shared `device.py` helpers, defaulting to `False`, and only `wpas.py` passes `True`. `_set_p2p_oper_channel` stays unconditional as you suggested. `session.py` is untouched, so the `nm` path keeps the cosmetic warning it had before.

Dropped the netdev block and replaced the comment with a short note on why it isn't needed - the distro already grants it on Debian, and the group doesn't exist on Arch. I also reworded the three other places that described that grant as ours: `_gdbus_call`'s docstring, the `--wfd-p2p-backend` help text, and the README / DOCUMENTATION entries.

One thing I ran into while threading the parameter through. `_p2p_device_iface_paths`'s `_priority` sort key reads `Ifname` via `_nm_get_string`, which is hardcoded to NetworkManager's destination, so it never resolved a `/fi/w1/wpa_supplicant1/...` path - it returns `""`, every candidate gets the same priority, and `sorted()` just preserves whatever order the supplicant listed. So the `p2p-dev-<iface>` preference has been inert since #107. `device.py` can't import `wpas.py`'s `_wpas_get_property` (wpas.py imports device.py), so I moved that helper down to `dbus.py` and both use it now. Say the word if you'd rather I split that out into its own PR and keep this one to the review.

On tests: the two existing ones now pin the wpas path explicitly, and I added ones asserting `_p2p_device_iface_paths` and `_set_p2p_device_name` do *not* escalate at their defaults - that's the #113 case specifically. 97 tests, with the same single pre-existing `pystray` import failure in my sandbox.

IlyaP358 (approved):
Both look right, thanks. I measured the nm path again and it now makes zero sudo calls where it made two before, and `session.py` is untouched, so #113 keeps its cosmetic warning.

On the sort key, good catch, and I confirmed it on my own machine. `_nm_get_string` returns `''` for a `/fi/w1/wpa_supplicant1/...` path while `_wpas_get_string` returns `wlan0`, so the `p2p-dev-` preference really has been inert since #107. Keep it in this PR, it is small and it is in the function you had to touch anyway. I would rather not make a habit of it, but splitting this one would just cost another round.

Merged cleanly against current dev, 123 tests pass there, and the policy file comes out with `wheel` and `sudo` and no stray netdev warning. 

Merging! =]

PR #111, opened 2026-09-02
Add Spanish (es) translation

Jemo121:
## Summary

Adds Spanish (`es`) to `src/i18n/translations.json` per #94 - all 29 strings, plus the README credits table.

Register is Mexican Spanish throughout: "computadora"-style vocabulary, infinitive/impersonal phrasing for menu commands (e.g. "Detener transmisión", "Volver a buscar dispositivos") to match how the existing German and French entries handle the same commands.

Not closing #94 with this - German and French are already in, and that issue is the standing invitation for whichever languages are still missing (Italian, etc.), so it should stay open past this PR.

## Testing

- `python3 -m json.tool` - valid JSON, and diffed to confirm only new `"es"` keys were added (no existing `en`/`ru`/`cs`/`da`/`de`/`fr` values touched).
- Verified `LANG=es_MX.UTF-8` resolves to `es` through `i18n.py`'s own locale-suffix stripping and returns the right strings:

```
FLUXCAST_LANG=es_MX.UTF-8 python3 -c "..."
resolved lang: es
'Stop Casting' -> 'Detener transmisión'
```

- Full test suite still green (91 tests, one pre-existing unrelated failure from a missing `pystray` import in this sandbox).

Issue #94, opened 2026-08-24
📢 Help wanted: Looking for translators (German, French, Spanish, Italian, etc.)


IlyaP358:
**FluxCast now supports easy localization**, and we want to expand beyond English, Russian, and Czech! 

If you want to see the system tray interface in your native language, you don't need to write any Python code. 

### How to contribute a translation:
1. Open `src/i18n/translations.json`.
2. Add your language code (e.g., `"de"`, `"es"`, `"fr"`) and the translated text to the existing keys.
3. Open a Pull Request!

It takes less than 5 minutes as there are only ~30 strings to translate. 

Thank you for making FluxCast better!

therealresonix:
Please switch to Weblate or something to make it easier in the long term.

PR #112, opened 2026-09-02
Add Brazilian Portuguese (pt) translation

Jemo121:
## Summary

Adds Brazilian Portuguese (`pt`) to `src/i18n/translations.json` per #94 - all 29 strings, plus the README credits table.

Calling this out explicitly as **Brazilian** Portuguese, not European: the i18n system only has one `pt` slot, and the two variants differ enough in everyday vocabulary (e.g. "tela" vs "ecrã" for screen) that it's worth being precise about which one this is, both in the README credit line and here.

Register matches the existing German/French/Spanish entries: infinitive/impersonal phrasing for menu commands ("Parar transmissão", "Procurar dispositivos novamente").

## Testing

Same checks as #111 (the Spanish PR):
- Valid JSON, diffed to confirm only new `"pt"` keys were added - no existing language values touched.
- `LANG=pt_BR.UTF-8` resolves to `pt` through `i18n.py`'s locale-suffix stripping and returns the right strings.
- Full test suite still green (91 tests, one pre-existing unrelated failure from a missing `pystray` import in this sandbox).

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ The tray picks your language up from the system locale, no configuration needed.
 | Danish | [normann](https://github.com/peternormann) |
 | Deutsch | [therealresonix](https://github.com/therealresonix) |
 | Français | [t0mab](https://github.com/t0mab) |
+| Português (Brasil) | [Jemo121](https://github.com/Jemo121) |
 
 Adding your language means editing one JSON file and no code at all.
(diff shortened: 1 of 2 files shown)


IlyaP358 (approved):
lgtm =] thanks, merging

PR #115, opened 2026-09-04
fix: bound HLS playlist with delete_segments (fixes #110)

### LINKED ISSUE #110: HLS segments are never deleted, so a long DLNA/Cast session fills /tmp (tmpfs)
## Problem

The Cast (`--protocol cast`) HLS output and the Cast metadata disagree about what kind of stream this is.

`play_media()` (via `pychromecast`, `src/cast/cast.py`) sends the media load request with `stream_type` left at its default, `STREAM_TYPE_LIVE` - telling the receiver "this is a live broadcast, seek to the live edge, tolerate irregular segment timing."

But both places FluxCast actually builds the HLS manifest use `hls_flags=append_list`:

```
src/capture/capture.py:292:        "-p", "hls_flags=append_list",
src/capture/capture.py:551:        "-hls_flags", "append_list",
```

`append_list` disables FFmpeg's sliding window entirely - segments are appended to the playlist forever and never removed. That's FFmpeg's mechanism for producing an `EVENT`-type playlist (a growing, seekable-from-the-start recording), which is a fundamentally different HLS semantic than a live stream. A live HLS playlist should have no `#EXT-X-PLAYLIST-TYPE` tag at all and drop old segments as new ones arrive, which is what signals "live" to a receiver in the first place.

So the request tells the Cast receiver "treat this as live" while the manifest itself is structured like a growing VOD-ish recording. A strict HLS/Cast implementation could reasonably choke on that inconsistency - which would explain reliability reports concentrated on specific TV vendors (e.g. #2) rather than everyone, since receivers vary a lot in how strictly they validate this. It also means the manifest grows unbounded for the whole session, which is wasted parse cost on every manifest refresh regardless of whether it ever causes a hard failure.

## Not the same as #51

#51 ("Various Chromecast Issues") covered similar-sounding symptoms and was closed with a different, unrelated diagnosis - real-time H.264 encoding being CPU-bound on low-power hardware. That's a real, separate constraint. This issue is about the HLS manifest type contradicting the Cast stream type regardless of CPU headroom - checked, and neither that issue nor any other on the tracker mentions `append_list`, `hls_flags`, `stream_type`, or playlist type.

## Proposed fix

Swap `hls_flags=append_list` for `hls_flags=delete_segments` (removes segment files as they roll off the window, keeping a bounded sliding playlist) at both call sites, and confirm no `#EXT-X-PLAYLIST-TYPE` tag ends up in the generated manifest - so the playlist's own semantics actually match the `LIVE` stream type already being sent to the receiver, instead of contradicting it.

Happy to put together the PR if this is the right read of it.

v01dst:
## Problem
Fixes #110 — long DLNA/Cast sessions fill /tmp (tmpfs) because HLS segments are never deleted.

Both capture paths build an HLS manifest with `hls_flags=append_list`:
- `src/capture/capture.py:292` (wf-recorder Wayland path, `hls_list_size=3` set but ignored)
- `src/capture/capture.py:551` (X11 x11grab path, `hls_list_size=6` set but ignored)

## Root cause
`append_list` disables FFmpeg sliding-window deletion entirely — segments append forever, producing EVENT-style growing playlist. This contradicts the Cast path which sends `STREAM_TYPE_LIVE` in `src/cast/cast.py`, and wastes parse cost on every manifest refresh. Receivers that validate strictly can choke, matching vendor-specific reliability reports.

## Fix
- `src/capture/capture.py:292`: `hls_flags=append_list` → `hls_flags=delete_segments`
- `src/capture/capture.py:551`: `-hls_flags append_list` → `-hls_flags delete_segments`

Existing `hls_time` / `hls_list_size` values now take effect as bounded sliding windows (3 and 6 segments), segment files roll off disk, no EXT-X-PLAYLIST-TYPE tag is emitted so playlist semantics match LIVE.

Minimal 2-line diff, no refactoring.

## Verification
- Repro: `grep append_list src/capture/capture.py` showed 2 hits before, 0 after; `delete_segments` 0 before, 2 after.
- `python3 -m py_compile src/capture/capture.py` → OK
- `python3 -m unittest tests.test_wf_recorder -v` → 4 tests OK
- `git diff --stat` → 1 file, 2 insertions, 2 deletions

### DIFF
--- a/src/capture/capture.py
+++ b/src/capture/capture.py
@@ -289,7 +289,7 @@ def _start_capture_wf_recorder(
         "-p", "tune=zerolatency",
         "-p", "hls_time=1",
         "-p", "hls_list_size=3",
-        "-p", "hls_flags=append_list",
+        "-p", "hls_flags=delete_segments",
         "-p", "pix_fmt=yuv420p",
         "-p", "profile=main",
         "-f", "/tmp/fluxcast/stream.m3u8",
@@ -548,7 +548,7 @@ def _start_capture_x11grab(
         "-f", "hls",
         "-hls_time", "2",
         "-hls_list_size", "6",
-        "-hls_flags", "append_list",
+        "-hls_flags", "delete_segments",
         "-hls_segment_filename", os.path.join(hls_dir, "stream%d.ts"),
         os.path.join(hls_dir, "stream.m3u8"),
     ]


IlyaP358 (approved):
lgtm👍 thanks for the quick fix =]

PR #118, opened 2026-09-07
Accept wfd-aosp-pmt-pid, wfd-uibc and wfd-go-intent in the tray config

### LINKED ISSUE #116: Tray config rejects flags some sinks require, and --wfd-aosp-pmt-pid is undocumented
The tray builds its launch command from `~/.config/fluxcast/config`, but `_MODE_OPTIONS["wfd"]` in `src/ui/tray_config.py` only whitelists 11 of the 22 `--wfd-*` flags. Anything else is dropped:

```
[FluxCast tray config] WARNING: unknown option 'wfd-aosp-pmt-pid' in [wfd]; ignoring it
```

That warning goes to stderr, and the tray sends stderr to its log file, so the user never sees it. They just get a cast that doesn't work, with no visible reason.

Three of the missing flags are worth adding, all of them stable per-setup properties rather than debugging switches:

- `wfd-aosp-pmt-pid` required by AOSP-based receivers, so those users can't use the tray at all today (see #84)
- `wfd-uibc` input back channel is a normal feature, not a one-off
- `wfd-go-intent` some sinks need a value other than the default

The rest should stay out. `wfd-p2p-backend` and `wfd-p2p-channel` need sudo for the IP layer and the tray has no terminal to prompt in; `wfd-dump-ts` and `wfd-ffmpeg-stats` are one-off debugging; `wfd-peer` and `wfd-monitor` are already supplied by the tray itself.

Separately, `--wfd-aosp-pmt-pid` doesn't appear in DOCUMENTATION.md or the README at all, so there is no way to discover it other than reading `--help`

7487:
Fixes #116

`_MODE_OPTIONS["wfd"]` in `src/ui/tray_config.py` only whitelisted 11 of the `--wfd-*` flags, so `wfd-aosp-pmt-pid = true` in `~/.config/fluxcast/config` was dropped with a warning that only reaches the tray log. AOSP-sink users got a black screen from the tray with no visible reason.

Changes:

- `src/ui/tray_config.py`: accept `wfd-aosp-pmt-pid` (flag), `wfd-uibc` (flag) and `wfd-go-intent`. Go intent is validated against the CLI's `0-15` range with a dedicated `_go_intent` validator, since `_positive_int` would reject the default `0`.
- Left out, as the issue suggests: `wfd-p2p-backend`/`wfd-p2p-channel` (may need sudo, no terminal to prompt in), `wfd-dump-ts`/`wfd-ffmpeg-stats` (one-off debugging), `wfd-peer`/`wfd-monitor` (supplied by the tray).
- `tests/test_tray_config.py`: covers the three new keys being forwarded (including `wfd-go-intent = 0` and a false flag being omitted) and an out-of-range go intent warning and falling back.
- `documentation/DOCUMENTATION.md`: `--wfd-aosp-pmt-pid` added to the WFD flag list, given a Flag Details entry under WFD Media (what it changes in the TS, when to use it, ffmpeg-only muxing per `pipeline.py`), and the three keys added to the tray config option list.
- `README.md`: short usage example for Android-based sinks next to the other troubleshooting snippets.

Test run (`PYSTRAY_BACKEND=dummy python -m unittest tests.test_tray_config`): 12 tests, OK.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,13 @@ python3 src/main.py --protocol wfd --wfd-capture-backend wf-recorder
 python3 src/main.py --protocol wfd --wfd-capture-backend x11grab
 ```
 
+Android-based sinks (TV boxes, projectors) that connect but stay on a black
+screen:
+
+```bash
+python3 src/main.py --protocol wfd --wfd-aosp-pmt-pid
+```
+
 WFD P2P connection troubleshooting - talk to wpa_supplicant directly instead
 of going through NetworkManager, and optionally pin the group to a 2.4GHz
 channel for sinks that don't support Wi-Fi Direct on 5GHz:
--- a/documentation/DOCUMENTATION.md
+++ b/documentation/DOCUMENTATION.md
@@ -61,6 +61,7 @@ python3 src/main.py --protocol cast
 - `--wfd-capture-backend auto|portal|wf-recorder|x11grab|gst-x11`
   - `auto` uses `portal` first on KDE/GNOME Wayland, then `wf-recorder` fallback.
 - `--wfd-latency-log [PATH]`
+- `--wfd-aosp-pmt-pid` write MPEG-TS tables the way Android (AOSP) sinks expect them
 - `--wfd-no-audio`
 - `--wfd-audio-device DEVICE`
 - `--wfd-rtsp-port PORT`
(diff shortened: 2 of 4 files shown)


IlyaP358 (approved):
lgtm 👍. merging

PR #119, opened 2026-09-07
Give firewalld queries the Polkit budget; probe state via systemd

### LINKED ISSUE #114: firewalld query times out before Polkit can be answered, so the RTSP port stays closed
Reported by @Tenyama on #113, by @Zosoled on #53 before that, and once more by a user off-issue. On systems where firewalld gates `firewall-cmd` through Polkit, the read-only query FluxCast runs before opening the port dies at 3 seconds, long before the dialog can be answered:

```
[FluxCast WFD] Could not open firewalld port 7236/tcp automatically (could not check existing rule: Command '['firewall-cmd', '--zone=nm-shared', '--query-port=7236/tcp']' timed out after 3.0 seconds).
```

The port then stays closed, the TV's connection back to us on 7236 never arrives, and the session sits in "waiting for TV RTSP/WFD session" until Ctrl+C. Nothing tells the user the firewall is the reason.

Not a duplicate of #76. That one asked for a `--query-port` pre-check and, on strict Polkit setups, a bail-out with the `--wfd-no-firewall` hint instead of more prompts, and that is exactly what shipped. This issue is about that bail-out being a dead end: those users can't cast at all, and three of them have now reported it as a connection problem rather than a firewall one.

#76 also assumed that only `--state` is exempt from Polkit on such systems. That doesn't hold everywhere. On my own machine the doctor times out on `--state` itself, and so did another user's, so the exemption can't be relied on.

Three call sites still use 3 seconds:

- `src/wfd/firewall.py:18` `_FIREWALL_QUERY_TIMEOUT`, used at `:48` for `--zone=nm-shared --query-port`. This is the one in the log above.
- `src/diagnostics.py:674` `firewall-cmd --state`
- `src/diagnostics.py:701` `firewall-cmd --query-port`

The two `--query-port` calls have no non-Polkit equivalent, so they should get the same budget as the authorization call they precede. The doctor's `--state` does have one: `_firewalld_active()` in `src/wfd/firewall.py:11` already answers the same question with `systemctl is-active firewalld`, which never touches Polkit. Using that in `_firewalld_check()` would also stop the doctor from raising a dialog on every startup.

7487:
Fixes #114

On Polkit-gated firewalld hosts the read-only `--query-port` run before `--add-port` died at 3 s, before the dialog could be answered, so the port stayed closed and the session sat in "waiting for TV RTSP/WFD session". The doctor's `--state` had the same 3 s limit and raised a Polkit dialog on every startup.

- `src/wfd/firewall.py`: the session-path `--query-port` now uses `_FIREWALL_AUTH_TIMEOUT` (60 s), the same budget as the `--add-port` it precedes. `_FIREWALL_QUERY_TIMEOUT` is removed (also from the `wfd/__init__.py` re-export; nothing else referenced it).
- `src/diagnostics.py`: `_firewalld_check()` replaces `firewall-cmd --state` with `_firewalld_active()` (`systemctl is-active firewalld`), which never touches Polkit, so `--doctor` no longer raises a dialog on every startup. `_firewalld_active()` is tri-state: `True`/`False` when systemd answered, `None` when it could not be asked (no `systemctl`, no systemd, timeout), which the doctor reports as "could not verify" rather than "not running". The doctor's informational `--query-port` keeps the 3 s probe budget so session startup never waits on a Polkit dialog; a timeout is reported as "could not verify".

`_firewalld_active()` moved from `wfd/firewall.py` into `diagnostics.py` rather than being imported the other way: `wfd/__init__.py` already does `from diagnostics import ...`, so `diagnostics` importing `wfd.firewall` is a circular import. `wfd/firewall.py` imports it from `diagnostics` and treats `None` like inactive (nothing to open), so `wfd._firewalld_active` still resolves and the existing `patch_all("_firewalld_active", ...)` tests keep working. `_FIREWALL_AUTH_TIMEOUT` stays in `wfd/firewall.py`.

Tests: `test_diagnostics` covers the doctor verdict for each `systemctl` outcome (active / inactive / no systemd / missing / hang) with the real `_firewalld_active`, that a not-running or unverifiable firewalld never invokes `firewall-cmd`, that the doctor's query passes the 3 s budget, and that a query timeout is "could not verify". `test_wfd` asserts the session query and add calls share `_FIREWALL_AUTH_TIMEOUT` and that `None` skips `firewall-cmd`. `python3 -m unittest tests.test_diagnostics tests.test_wfd`: 65 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/diagnostics.py
+++ b/src/diagnostics.py
@@ -666,44 +666,43 @@ def _ufw_check() -> Optional[Check]:
     )
 
 
+# firewalld gates every firewall-cmd call through Polkit, read-only queries
+# included, so on some hosts `--query-port` blocks until the user answers a
+# dialog. Give it the same budget as `--add-port` rather than a 3 s probe.
+_FIREWALL_AUTH_TIMEOUT = 60.0
+
+
+def _firewalld_active() -> bool:
+    # `firewall-cmd --state` goes through Polkit too; asking systemd does not.
+    if not shutil.which("firewall-cmd"):
+        return False
+    try:
+        result = _run(["systemctl", "is-active", "firewalld"], timeout=3.0)
+    except (OSError, subprocess.TimeoutExpired):
+        return False
+    return result.returncode == 0 and result.stdout.strip() == "active"
+
+
 def _firewalld_check() -> Optional[Check]:
     if not shutil.which("firewall-cmd"):
         return None
 
-    try:
-        state = _run(["firewall-cmd", "--state"], timeout=3.0)
-    except (OSError, subprocess.TimeoutExpired) as exc:
-        return Check("firewall (firewalld)", STATUS_WARN, "could not query firewalld state", str(exc))
-
-    # firewalld gates firewall-cmd through polkit, so on a headless/gated host
-    # `--state` and `--query-port` can fail with an authorization error rather
-    # than a real answer. Trust only what firewall-cmd literally prints: a
-    # non-zero exit is not proof the firewall is down, so an auth failure must
-    # never become a definitive OK-or-closed. Report "couldn't verify" instead.
-    state_out = state.stdout.strip()
-    state_all = (state.stdout + state.stderr).strip()
-    if state_out != "running":
-        if "not running" in state_all.lower():
-            return Check(
-                "firewall (firewalld)",
-                STATUS_OK,
-                "firewalld is not running; port not blocked",
-                state_all,
-            )
+    if not _firewalld_active():
         return Check(
             "firewall (firewalld)",
-            STATUS_WARN,
-            "could not verify firewalld state (firewall-cmd did not report running/not running)",
-            state_all,
+            STATUS_OK,
+            "firewalld is not running; port not blocked",
         )
 
     try:
-        query = _run(["firewall-cmd", f"--query-port={WFD_RTSP_PORT}/tcp"], timeout=3.0)
+        query = _run(["firewall-cmd", f"--query-port={WFD_RTSP_PORT}/tcp"], timeout=_FIREWALL_AUTH_TIMEOUT)
     except (OSError, subprocess.TimeoutExpired) as exc:
         return Check("firewall (firewalld)", STATUS_WARN, "could not query firewalld port", str(exc))
 
     # `--query-port` prints `yes`/`no` (exit 0/1) for a real answer; anything
     # else — empty output, an auth error — means we could not check the port.
+    # A non-zero exit is not proof the port is closed, so an auth failure must
+    # never become a definitive OK-or-closed. Report "couldn't verify" instead.
     query_out = query.stdout.strip()
     if query_out == "yes":
         return Check(
--- a/src/diagnostics.py
+++ b/src/diagnostics.py
@@ -666,38 +666,59 @@ def _ufw_check() -> Optional[Check]:
     )
 
 
-# firewalld gates every firewall-cmd call through Polkit, read-only queries
-# included, so on some hosts `--query-port` blocks until the user answers a
-# dialog. Give it the same budget as `--add-port` rather than a 3 s probe.
-_FIREWALL_AUTH_TIMEOUT = 60.0
+def _firewalld_active() -> Optional[bool]:
+    """True/False when systemd reports firewalld active/inactive, None when it
+    could not be asked (no systemctl, no systemd, timeout).
 
-
-def _firewalld_active() -> bool:
-    # `firewall-cmd --state` goes through Polkit too; asking systemd does not.
+    `firewall-cmd --state` goes through Polkit too; asking systemd does not.
+    """
     if not shutil.which("firewall-cmd"):
         return False
     try:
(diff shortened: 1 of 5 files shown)


IlyaP358 (changes requested):
Thanks, and the main fix works. I tested it on a real firewalld host with a Polkit agent running. the dialog appeared, I entered the password, and the port opened in 13.9 seconds. On `dev` the same call dies at 3 seconds, so this closes the actual bug. I also checked the import move you described, `import diagnostics`, `import wfd`, `import wfd.firewall` and `import main` all work in any order, and the `wfd._firewalld_active` re-export still resolves so the existing tests keep patching it.

Two things I'd like changed before merging.

**The doctor now gives a definite answer when it doesn't have one.** `_firewalld_active()` returns `False` both when systemd says firewalld is inactive and when it couldn't ask at all, so the check reports `ok firewalld is not running; port not blocked` in both cases. I mocked the four possibilities:

```
systemctl: active      -> [ok] firewalld is running and allows port 7236
systemctl: inactive    -> [ok] firewalld is not running; port not blocked
systemctl missing      -> [ok] firewalld is not running; port not blocked
systemctl hangs        -> [ok] firewalld is not running; port not blocked
```

The last two are wrong. On a host without systemd, firewalld can be running and blocking while we tell the user it isn't. The code you removed guarded against exactly this, its comment said an auth failure must never become a definitive OK-or-closed. Could `_firewalld_active` distinguish "systemd said inactive" from "couldn't ask", so the doctor can report "couldn't verify" for the second case? `firewall.py` can keep treating both as "don't try".

**The doctor's query doesn't need the auth budget.** `run_diagnostics()` runs at the start of every WFD session, so the 60 second timeout on its informational `--query-port` means a startup can sit and wait on a Polkit dialog. I measured 25 seconds on my machine, which is D-Bus giving up on its own reply timeout rather than our 60. The session path in `firewall.py` should keep the long budget since the port actually matters there and the user is expecting a prompt, but the doctor should stay short and just say it couldn't verify.

Everything else looks good, and the tests you added are the right ones, especially asserting that a not-running firewalld never invokes `firewall-cmd`.

7487:
Thanks for testing on a real host. Both points addressed in 93f6b12.

**Tri-state.** `_firewalld_active()` now returns `True` / `False` only when systemd actually answered (`active`, or `inactive`/`failed`), and `None` when it could not be asked: `systemctl` missing, timeout, or an empty reply such as "System has not been booted with systemd as init system". The doctor reports `None` as `[warn] could not verify whether firewalld is running`; `firewall.py` is unchanged apart from a comment, since `not None` already means "don't try". Your table now reads:

```
systemctl: active      -> [ok]   firewalld is running and allows port 7236
systemctl: inactive    -> [ok]   firewalld is not running; port not blocked
no systemd (PID 1)     -> [warn] could not verify whether firewalld is running (systemctl gave no answer)
systemctl missing      -> [warn] could not verify whether firewalld is running (systemctl gave no answer)
systemctl hangs        -> [warn] could not verify whether firewalld is running (systemctl gave no answer)
```

`firewall-cmd` is only invoked in the first row.

**Doctor query budget.** The doctor's `--query-port` is back on the 3 s probe budget, and a timeout there is reported as `[warn] could not verify whether firewalld allows port 7236/tcp` rather than blocking startup. `_FIREWALL_AUTH_TIMEOUT` moved back to `wfd/firewall.py` since the session path is now its only user; the `wfd._FIREWALL_AUTH_TIMEOUT` re-export is unchanged.

Tests: `test_firewalld_doctor_reports_each_systemctl_outcome` runs the real `_firewalld_active` through those five outcomes and asserts status, message, and whether `firewall-cmd` was invoked; `test_firewalld_doctor_query_keeps_short_budget` asserts the doctor passes 3.0; `test_firewalld_doctor_query_timeout_is_could_not_verify` covers the query hang. In `test_wfd`, the session path now asserts the query still gets `_FIREWALL_AUTH_TIMEOUT`, plus a case that `None` skips `firewall-cmd` entirely. `python3 -m unittest tests.test_diagnostics tests.test_wfd`: 65 passed.

IlyaP358 (approved):
Both points look good, thanks for the quick turnaround.

I verified the tri-state myself against the real `_firewalld_active`, all six outcomes match what you described, and `firewall-cmd` is only invoked when systemd actually says active. The session path kept the 60 second budget on both the query and the add, which is exactly the split I wanted, and the doctor is back to a 3 second probe. No import cycles in any order, the `wfd._FIREWALL_AUTH_TIMEOUT` re-export still resolves, and the full suite passes at 115 tests.

Earlier I confirmed the original bug is fixed on a real firewalld host: the Polkit dialog appeared, I entered the password, and the port opened in 13.9 seconds where `dev` gives up at 3. That was what three users kept hitting, so this closes it properly.

Merging 👍
 Thank you very much for your help. You'll be on [https://fluxcast.dev/contributors.html](https://fluxcast.dev/contributors.html) =]

Issue #113, opened 2026-09-02
fluxcast-install-system: command not found


Tenyama:
## Note

- Installation method: PyPi
- `fluxcast` runs normally but cannot cast to LG TV (most likely my error) (output below)

## System

- EndeavourOS (Linux 7.1.10-arch1-1)
- Hyprland (0.56.2)
- zsh 5.9.2
- GPU : Quadro RTX 3000 Mobile / Max-Q @  [Discrete]
- GPU : Intel UHD Graphics @ 1.15 GHz [Integrated]

## Output of `fluxcast`

```
FluxCast Doctor] System capability report
[FluxCast Doctor] Miracast/WFD looks possible via NetworkManager; raw supplicant access is optional for this backend.

  Status Check                  Details
  ------ ---------------------- ------------------------------------------
  ok     python                 runtime (3.13.13 on Linux 7.1.10-arch1-1)
  ok     ffmpeg                 video/audio transcoding (/usr/bin/ffmpeg)
  warn   wf-recorder            Wayland/wlroots screen capture (not found in PATH)
  ok     xdg-desktop-portal     Wayland portal stack is running (xdg-desktop-portal-hyprland)
  ok     pactl                  PulseAudio/PipeWire-Pulse audio monitor detection (/usr/bin/pactl)
  ok     xrandr                 X11 monitor detection fallback (/usr/bin/xrandr)
  ok     nmcli                  NetworkManager Wi-Fi Direct control (/usr/bin/nmcli)
  ok     dnsmasq                NetworkManager P2P DHCP server (/usr/bin/dnsmasq)
  ok     iw                     kernel Wi-Fi interface inspection (/usr/bin/iw)
  ok     wpa_cli                active Wi-Fi Direct scan/control (/usr/bin/wpa_cli)
  ok     gdbus                  passive wpa_supplicant D-Bus capability checks (/usr/bin/gdbus)
  ok     gst-launch-1.0         optional future WFD GStreamer pipeline (/usr/bin/gst-launch-1.0)
  ok     gst-inspect-1.0        optional future WFD codec inspection (/usr/bin/gst-inspect-1.0)
  ok     gst element pipewiresrc Wayland portal GStreamer capture source (pipewiresrc)
  ok     gst element x264enc    Wayland portal GStreamer H.264 encoder (x264enc)
  ok     dbus_next              WFD portal capture control plane for KDE/GNOME Wayland (python module is installed)
[...]

Issue #109, opened 2026-09-02
Debian/Ubuntu packaging (.deb) - no apt-installable path today


Jemo121:
## Problem

There's no apt-installable path today. Current options are AppImage, PyPI (`pip install fluxcast` + a separate `fluxcast-install-system` step), the AUR, or from source. That's a real gap for the Debian/Ubuntu userbase specifically - "pip install a system tray app, then run a second privileged script" is a lot more friction than `apt install fluxcast` on a platform that's used to apt-only workflows.

## Proposed first step

A downloadable `.deb` attached to GitHub Releases, mirroring the AppImage that's already there. Not proposing a hosted PPA or an official Debian archive submission as a starting point - both are a much bigger, ongoing commitment (archive review, dependency bumps and changelog entries every release) than a one-off release artifact, and that's a call for whoever's going to maintain it, not something to just show up with in a PR.

Mechanically this is mostly `debhelper` boilerplate:
- `debian/control`, `rules`, `changelog`, `copyright`
- `postinst`/`postrm` to install `meta/zz-dev.fluxcast.wpa-supplicant.conf` as a proper conffile and reload dbus, instead of the manual step `meta/install.sh` does today
- a `.desktop` entry + icon cache update

## Known friction

Some Python deps (`pystray`, `pychromecast`) don't have clean Debian packages, so a strict "Debian-policy-pure" package isn't realistic without vendoring. The practical option is a private venv under something like `/opt/fluxcast`, similar in spirit to what the PyPI/AppImage paths already do - not maximally elegant, but it's what actually ships and gets maintained by third-party apt-distributed tools in practice.

## Ask

Mainly want to check whether this is wanted before building it - happy to put together the `.deb` and open a PR if so, but a proper packaging setup is ongoing surface area (version bumps, testing installs/upgrades, conffile handling), not a fire-and-forget contribution.

IlyaP358:
Yes, this is wanted, go ahead =]

Your scoping is right too. A `.deb` attached to Releases is the artifact I actually want now, and I'd rather not sign up for a PPA or an archive submission yet. The venv under `/opt/fluxcast` is fine for that artifact, and #113 shows the current PyPI path is worse in practice than the README claims, so this genuinely removes friction rather than adding a fourth way to install the same thing.

[...]

PR #122, opened 2026-09-10
Debian/Ubuntu packaging: .deb built per release for GitHub Releases

Jemo121:
## Summary

Adds a `.deb` build for Debian and Ubuntu, attached to GitHub Releases alongside the AppImage. Closes the gap in #109: there was no apt-installable path, only AppImage, PyPI plus a separate privileged step, the AUR, or from source.

Scoped as you asked in #109 - a release artifact, not a PPA and not an archive submission. Nothing here signs us up for archive review or per-release changelog obligations.

## How it stays single-sourced

`debian/rules` runs `meta/install.sh` with `DESTDIR` rather than restating the file list, the `/opt` layout or the D-Bus policy path. It is the same script a from-source user runs, so the three install paths cannot drift. The build asserts this: the installed policy is byte-identical to `meta/zz-dev.fluxcast.wpa-supplicant.conf` and owned by dpkg.

`postinst` reloads dbus, so Wi-Fi Direct works straight after `apt install` with no second privileged command. That is the friction #113 ran into on the PyPI path.

## The upgrade case

`meta/install.sh` and the AppImage first-run setup both write into `/opt/fluxcast` and `/usr/share/dbus-1/system.d` without dpkg knowing. dpkg overwrites the files it also ships, but modules only an older layout had would survive and shadow the packaged ones on `sys.path`.

`preinst` detects a `/opt/fluxcast` that dpkg does not own and clears it - nothing under there is user data. It also removes the venv on upgrade, since it ships whole and is unpacked immediately afterwards. I tested this by planting a stale module before installing over a source install; it is gone afterwards on every target.

## One .deb per release, not one overall

This is the part worth your review.

The bundled virtualenv holds wheels ABI-tagged for the build host's Python minor version, and its `bin/python3` is a symlink to the system interpreter. A bookworm build installed on trixie therefore resolves to a `site-packages` directory that is not there and loses every dependency - it fails at first import, not at install time. I hit this before pinning it.

So `dh_gencontrol` pins each build to the `python3` it was built against, and CI produces one artifact per release with a `~codename` version suffix:

| Build | Release | Python |
|-------|---------|--------|
| `~bookworm` | Debian 12 | 3.11 |
| `~trixie`   | Debian 13 | 3.13 |
| `~noble`    | Ubuntu 24.04 | 3.12 |
| `~resolute` | Ubuntu 26.04 | 3.14 |

Installing the wrong one is refused by apt rather than half-working. The alternative - bundling a standalone Python so one artifact covers everything - would roughly double the download for every user, so I did not take it.

## Dependencies

`Depends` is limited to what FluxCast cannot start or cast without: the Python bound, the GTK/AppIndicator introspection data, dbus, NetworkManager, wpa_supplicant, iw and ffmpeg. The capture and transcode stacks (GStreamer, PipeWire, wf-recorder, pactl) are `Recommends`, and the plugin sets a given setup may never touch are `Suggests`. apt still pulls Recommends by default, so a plain `apt install` is unchanged; `--no-install-recommends` gives a genuinely minimal install rather than dragging in the whole media stack - about 47-53 MiB installed.

`gir1.2-ayatanaappindicator3-0.1` is a hard dependency - without it pystray cannot load a tray backend, which is the desktop entry's entire `Exec` line. My first build missed it and the tray failed on a clean install.

Python deps confirm your read in #109: `upnpclient` is absent from the archive entirely and `python3-pychromecast` is 9.4.0 against our 14.0.5 requirement, so the venv is the only workable option rather than a compromise.

## Launcher

`meta/fluxcast` now prefers the venv interpreter and falls back to `python3`, so one script serves both the package and a source install. I also quoted `"$@"` while in there - unquoted it dropped arguments containing spaces.

## Testing

Built and installed on **bookworm, trixie, noble and resolute**. On each: install over an existing from-source install, verify the stale tree is cleared and the policy is single-sourced, import every dependency plus the tray typelib from the venv, run `fluxcast --version`, reinstall to exercise the upgrade path, then remove and confirm `/opt/fluxcast` is gone.

I also installed each with `--no-install-recommends` to confirm the trimmed `Depends` really is sufficient: every dependency plus GTK and the tray typelib still import, and `--version`, `--help` and `--doctor` all run.

Cross-release installs were verified rejected in both directions (a noble deb on trixie, a trixie deb on noble, a bookworm deb on resolute, a resolute deb on bookworm).

Not tested against real casting hardware - this changes packaging only, no runtime code beyond the launcher.

## One thing I need from you

The `Maintainer:` field is a placeholder (`FluxCast Project <fluxcast@users.noreply.github.com>`). I did not want to put in an address that is not yours to give out. Tell me what it should be and I will set it.


Jemo121:
One policy question I should have raised in the description, since it decides how much this costs you over time.

Do we tie the build matrix to upstream support windows - once a release stops getting standard support from Debian or Ubuntu, we stop publishing a `.deb` for it? That bounds the ongoing surface area you flagged in #109: the matrix only ever holds supported releases, and dropping one is a one-line change rather than a judgement call each release.

Worth deciding now because the matrix is currently inconsistent about it, and I would rather that be your call than my accident:

- `debian:oldstable` and `debian:stable` are rolling tags. When Debian 14 ships they silently re-point, and we stop building for bookworm without anyone deciding to.
- `ubuntu:24.04` and `ubuntu:26.04` are pinned, so they never advance and will keep building for those releases past EOL until someone edits the file.

So right now Debian ages out by accident and Ubuntu never ages out at all.

If you want the EOL rule, I would pin all four explicitly and drop entries as releases go EOL, so every change to what we ship is visible in the diff. For reference: Ubuntu 24.04 is supported to June 2029, Ubuntu 26.04 to 2031, and Debian 12 to around June 2028 with LTS. Nothing in the current matrix is close to expiring, so this is about writing the rule down while it is cheap rather than fixing anything urgent.

Happy to add a short note to the README or the workflow recording whatever you decide.

IlyaP358:
Hi @Jemo121. Thanks, this is thorough work and it looks very solid!

I built it in containers on my side and installed it: the package builds, `fluxcast --version` reports `0.2.6~trixie`, the venv dependencies import, the D-Bus policy is owned by dpkg and byte-identical to `meta/`, the Python bound comes out as `python3 (>= 3.13), python3 (<< 3.14)`, and a trixie build really is refused on bookworm. I also checked the package contents: nothing from the repo leaks in beyond the app itself, no bytecode, no tests of ours. Reusing `meta/install.sh` from `debian/rules` is the right call, that is what keeps the three install paths from drifting.

I have since installed the noble build on a Mint vm. The tray applet and the CLI both work, which was the one thing containers could not show me.

Answering your two questions.

Maintainer: Please set it to `Illia Pukalov <illia@fluxcast.dev>` across the `debian/` metadata files.

On the matrix, yes, let us tie it to upstream support windows. Pin all four explicitly and drop an entry when a release goes EOL, so every change to what we publish shows up in the diff. You are right that it is inconsistent today: I checked, `debian:stable` currently resolves to trixie and `oldstable` to bookworm, so Debian would age out silently while Ubuntu never ages out at all. A short note in the workflow recording the rule would be good.

Three smaller things:

`Architecture: amd64` means no arm64 build. I am fine with that for now, but say so in the README so nobody goes looking.

The README table hardcodes `resolute` for Ubuntu 26.04. The build derives the codename from the image so the artifact is right either way, but please double check the table.

`--version` reports `0.2.6~trixie`, with the suffix. I like it, it tells me which build a bug report came from, just confirm that was deliberate.

Push those and I will merge =]

Jemo121:
Thanks for testing it that thoroughly, and for taking the noble build to a Mint VM - the tray was the one thing I could not check from containers either.

All four pushed.

**Maintainer.** Set to `Illia Pukalov <illia@fluxcast.dev>` in `debian/control` and the `debian/changelog` trailer. Those were the only two places carrying it.

**Matrix.** Pinned all four to explicit releases and wrote the rule into the workflow: publish while a release has upstream support, drop it when that ends.

Writing that down turned up something I had glossed over, so the scope is now stated properly. "Standard upstream support" promised more than the matrix delivers - Ubuntu interim releases have nine months of exactly that and never had an entry. The scope is Ubuntu LTS and Debian stable, and the README now says interim releases are unsupported instead of leaving it implied.

It also points them at the build that fits, because the gate turns out to be the `python3` minor version rather than the distribution. Ubuntu 25.10 ships python 3.13, so the `~trixie` build installs and imports there, while the `~bookworm` build is correctly refused - I checked both directions. Documented as a convenience and explicitly not a supported configuration, so nobody reads it as a fifth target. That also meant fixing the neighbouring paragraph, which said apt refuses "the wrong one" as though the distribution were the gate.

**arm64.** README now says the package is amd64 only, next to the note that a release leaves the table when its support does.

**The README codename table.** Checked, and it is correct - I ran each pinned tag and read `VERSION_CODENAME` from the image rather than trusting the table: `debian:12` bookworm, `debian:13` trixie, `ubuntu:24.04` noble, `ubuntu:26.04` resolute. Worth noting the table is still a hand-maintained mirror of something the build derives, so it can go stale without the artifacts being wrong. If that bothers you later, the honest fix is generating it, but I would not add that machinery for four rows.

**The `~codename` in `--version`.** Half deliberate, so let me be accurate rather than claim I planned it. The suffix on the *package version* is deliberate - it is what makes the four artifacts distinguishable and what apt sorts on. `--version` showing it follows from `debian/rules` stamping `__installed_version__` out of `dpkg-parsechangelog`, which was about not letting an installed copy report a git describe from the build host. I noticed the suffix came along, thought it was useful for exactly the reason you gave, and left it. So: intended to keep, not designed from the start.

**Re-verified after the changes**, since renaming the tags touches the build: both Debian targets rebuild, install over an existing from-source install, clear the stale tree, keep the policy byte-identical to `meta/`, survive an upgrade, remove cleanly, and still refuse a cross-release install. `Maintainer` and the per-release python bound are correct in the built packages (`0.1.5~bookworm` with `python3 (>= 3.11), (<< 3.12)`; `0.1.5~trixie` with `(>= 3.13), (<< 3.14)`).

One thing I did not touch: the changelog still says `0.1.5`, since CI rewrites it from the release tag. Your build reporting `0.2.6~trixie` is that working.

IlyaP358 (approved):
Excellent! Merging. Thanks for sorting out the metadata, stabilizing the matrix rules, and expanding the README coverage for interim releases. This is a massive step forward for making FluxCast easily installable.

On a personal note: mapping out the architecture for these updates with you has been incredibly smooth. You have a fantastic understanding of the Linux networking stack and the codebase. Since the project is getting a lot of traction lately, managing the incoming issues and PRs alongside my high school studies is becoming a bit heavy for a solo maintainer. 

If you're interested, I would love to invite you as an official Collaborator/Co-maintainer to help guide the project's architecture and triage issues together. No pressure at all, but let me know what you think!

Practically, this means collaborator access with write/push rights to help with issue triaging, PR reviews, and code architectural alignment (your review on the recent IWD PR was stellar, by the way =]). To keep things clean and predictable, I would keep cutting and publishing the final releases myself for now, with major features still routing through review cycles.

Jemo121:
Thank you — genuinely, that means a lot coming from you, and I'm glad the architecture conversations have been as useful on your end as they've been on mine.

I'll say yes, with one honest caveat about scope.

The caveat first: I've been around this project about five weeks. I know the WFD/P2P stack and the packaging well at this point, but I've barely touched capture, RTSP, DLNA or Cast. I'd rather not hold merge or admin rights over code I haven't read properly yet — so for now, think of me as a right hand rather than a second maintainer. Anything non-trivial I'd handle the way I handled #133 and #132: surface it, form an opinion, and leave the call to you. Releases stay yours, which is what you proposed anyway.

What I'm happy to take on starting now is the part that's actually eating your time: triaging incoming issues, first-pass reviews on PRs, and reproducing bug reports so they reach you already narrowed down. #135 is a good example — it's large and it lands in the stream path, and I can work through it properly before it needs your attention.

One thing worth saying plainly: I can't promise a response time. You're balancing this against school, and I'm balancing it against a full-time job, weekend work and university — so please don't hold anything waiting on me. If I'm slow, go ahead without me. I'd rather be useful than be a dependency.

Happy to revisit the scope in a few months, once I've spent real time in the parts of the codebase I don't know yet. If that works for you, send whatever access makes the triage practical and I'll start with the backlog.

---
Masters_Blessing

IlyaP358:
Welcome aboard! and thank you for laying out the scope yourself. Honestly it is the one I would have picked: knowing which parts of the codebase you have actually read is worth more to me than covering all of it.

Invite sent. One thing so it is not sitting in the back of your mind: main and dev now need a pull request and an approval from CODEOWNERS, which is me. So the merge button is not something you have to be disciplined about, it is simply not yours to press by accident. Releases stay with me, as we said.

Take #135, please. Triage and first-pass reviews are exactly where my evenings go, and what you did on #132 and #133 is already the pattern I want: dig in, form an opinion, hand me the call. That saved me a whole evening, and you also caught me being wrong on #125, which is worth more than a merge right.

On response times, do not give it a thought. FluxCast is something I do in the gaps between high school, it is not a job, and nothing here is urgent enough to wait on anyone. Go at whatever pace your week allows. If something genuinely cannot wait I will just say so and move on, no hard feelings in either direction.

Glad to have you here =]

Jemo121:
First triage pass done — the unlabelled backlog on open issues is clear.

I left `upstream bug` off #131 and `device-support` off #136; both would state a conclusion the evidence doesn't reach yet. Say the word if you read them differently.

One convention question: PRs here have never carried labels — the last twenty merged ones have none. Deliberate and issues-only, or would you like them labelled going forward?

IlyaP358:
Both calls are right, and for the right reason: neither label would have been a fact yet. Thanks for clearing the backlog, that is a real chunk of work.

On PRs, you are right that it is worth doing, and I checked before answering: eight of the last twenty merged PRs reference no issue at all, mostly translations, so for those a label is the only place that information could live. I would not do it by hand though, that is a tax on both of us forever. Path-based labelling with `actions/labeler` gets it for free: translations label themselves, packaging labels itself. I will set up the config and add the area labels for it. Keep the existing ones issues-only, they are all about state rather than area.

One more thing. We've been chatting under the .deb PR, which is merged and will eventually get buried. If you ever need to discuss architectural ideas or general questions, feel free to open a GitHub Discussion or pop into our [Discord](https://fluxcast.dev/#community). You can also always reach me directly at [illia@fluxcast.dev](mailto:illia@fluxcast.dev) or [illia.pukalov@teleinformatika.eu.](mailto:illia.pukalov@teleinformatika.eu.) Any of them works, whichever you prefer =]

Issue #121, opened 2026-09-10
NetworkManager Timeout


MarinManuel:
This does not seem directly related to fluxcast, but I'm hoping you have some insight.
For these logs, I have just printers next to me, but the output was the same when I was near a Miracast device.

```
./FluxCast-0.2.5-x86_64.appimage --protocol wfd --wfd-test-pattern --output-res 1280x720 --bitrate 3M 
[FluxCast Doctor] System capability report
[FluxCast Doctor] Miracast/WFD looks possible via NetworkManager; raw supplicant access is optional for this backend.

  Status Check                  Details
  ------ ---------------------- ------------------------------------------
  ok     python                 runtime (3.10.12 on Linux 7.0.0-30-generic)
  ok     ffmpeg                 video/audio transcoding (/tmp/.mount_FluxCakOOrtv/usr/bin/ffmpeg)
  warn   wf-recorder            wf-recorder was found but could not be executed (the AppImage wrapper needs the system wf-recorder package; install wf-recorder)
  ok     xdg-desktop-portal     Wayland portal stack is running (xdg-desktop-portal-kde)
  ok     pactl                  PulseAudio/PipeWire-Pulse audio monitor detection (/usr/bin/pactl)
  ok     xrandr                 X11 monitor detection fallback (/tmp/.mount_FluxCakOOrtv/usr/bin/xrandr)
  ok     nmcli                  NetworkManager Wi-Fi Direct control (/usr/bin/nmcli)
  ok     dnsmasq                NetworkManager P2P DHCP server (/usr/sbin/dnsmasq)
  ok     iw                     kernel Wi-Fi interface inspection (/usr/sbin/iw)
  ok     wpa_cli                active Wi-Fi Direct scan/control (/usr/sbin/wpa_cli)
  ok     gdbus                  passive wpa_supplicant D-Bus capability checks (/tmp/.mount_FluxCakOOrtv/usr/bin/gdbus)
  ok     gst-launch-1.0         optional future WFD GStreamer pipeline (/tmp/.mount_FluxCakOOrtv/usr/bin/gst-launch-1.0)
  ok     gst-inspect-1.0        optional future WFD codec inspection (/tmp/.mount_FluxCakOOrtv/usr/bin/gst-inspect-1.0)
[...]

PR #125, opened 2026-09-12
Only mark peers that advertised WFD data as Miracast sinks (#124)

### LINKED ISSUE #124: Peer list marks every Wi-Fi Direct device as WFD-capable
Reported indirectly by @MarinManuel on #121, whose scan showed two printers as valid sinks:

```
[0] 5E:3A:45:D2:2F:3B  DIRECT-3b-HP M227f LaserJet via NetworkManager
      WFD capability data detected
[1] AE:D5:64:82:B9:98  HL-L2350DW_BR3998 via NetworkManager
      WFD capability data detected
```

Neither is a Miracast sink. `print_scan` in `src/wfd/p2p/peers.py:162` prints that line when the peer's details contain `wfd_ies=`, and `src/wfd/p2p/nm.py:143` adds `wfd_ies=` whenever the raw gdbus output is truthy. For a device with no Wi-Fi Display data gdbus prints `(<@ay []>,)`, which is an empty array but a non-empty string, so the test always passes and every P2P peer gets the label.

The parsed value is already available one line earlier: `_parse_gdbus_byte_array` returns `[]` for that device and a byte list for a real sink. The check should use it.

This matters beyond cosmetics. A user who picks a printer gets a connection that sits in NetworkManager's `config` state until FluxCast times out after 35 seconds, with nothing explaining why, which is exactly what happened on #121.

Jemo121:
## Summary

Fixes #124. `_nm_scan` gated the `wfd_ies` detail on the raw gdbus string, but a peer with no Wi-Fi Display data still returns `(<@ay []>,)` - an empty array, but a non-empty string - so every P2P peer was reported as a valid sink. #121 scanned two printers and both were offered as Miracast targets.

The check now uses the parsed bytes, which `_parse_gdbus_byte_array` already produced one line earlier and which are `[]` for those devices.

## Reproduction

Replaying #121's two printers plus a real TV through `_nm_scan` and `print_scan`:

Before:

```
  [0] 5E:3A:45:D2:2F:3B  DIRECT-3b-HP M227f LaserJet via NetworkManager
      WFD capability data detected
  [1] AE:D5:64:82:B9:98  HL-L2350DW_BR3998 via NetworkManager
      WFD capability data detected
  [2] 16:7F:67:EB:4D:52  [LG] webOS TV OLED55A3PSA via NetworkManager
      WFD capability data detected
```

After:

```
  [0] 5E:3A:45:D2:2F:3B  DIRECT-3b-HP M227f LaserJet via NetworkManager
      no WFD capability data - probably not a Miracast sink
  [1] AE:D5:64:82:B9:98  HL-L2350DW_BR3998 via NetworkManager
      no WFD capability data - probably not a Miracast sink
  [2] 16:7F:67:EB:4D:52  [LG] webOS TV OLED55A3PSA via NetworkManager
      WFD capability data detected
```

## What else changed

Capability is carried on `WFDPeer` rather than re-derived by `print_scan` searching the `details` string, and peers known to have no WFD data now say so instead of printing nothing. Both go slightly beyond the one-line gate - I have left a comment below with the reasoning and am happy to cut either back if you would rather keep the diff tight.

## Testing

New `tests/test_peers.py`, 8 tests: the empty-array case from #124, a real sink's IEs, both wpa_cli paths, the unknown case, and `print_scan` output for all three states. I checked they fail when the fix is reverted rather than just passing alongside it.

Full suite goes 112 -> 120 with an unchanged failure set. The 6 that fail do so on pristine `dev` too - `pystray` and `dbus_next` are not installed in my environment, so `test_tray_config` and the `test_portal_window` cases cannot run. Nothing hardware-dependent here; the bug and the fix are both in string/byte parsing.


Jemo121:
Notes on the two things here that go beyond the one-line gate you described in #124. Both are easy to cut back if you would rather keep this tight.

**Capability lives on `WFDPeer` instead of being re-derived from `details`.** `print_scan` decided capability by searching the formatted details string, which is the same coupling that caused the bug - and it had two branches doing it, one per scanner. A `wfd_capable` field set by whichever scanner parsed the data removes the string matching entirely. The wpa_cli path sets it from `p2p_peer`'s own fields.

**Peers known not to advertise WFD now say so**, rather than printing nothing. That is aimed at the "nothing explaining why" part of #121 - silence is still ambiguous to someone who does not know a label is supposed to appear there.

One wrinkle worth flagging, since it is why the field is tri-state rather than a bool. On the wpa_cli path `p2p_peer` can fail or time out, leaving `details` empty. Treating that as "not capable" prints `probably not a Miracast sink` under a real TV and steers the user away from the device that would have worked - a false negative, which is a worse failure than the false positive being fixed here. So `wfd_capable` is `True` / `False` / `None`, where `None` prints nothing and matches today's behaviour. The NetworkManager path always knows, so it is only the wpa_cli fallback that can land on `None`.

I hit that one by accident: my first version used a plain bool, and a test with a timing-out `p2p_peer` lookup labelled a working LG TV as not a sink. There is now a regression test pinning it.

IlyaP358 (changes requested):
Checked this properly. The core fix is right, and the tests are honest: I reverted both source changes and the two that should fail do fail. Full suite is green here, 131 tests. Your 6 failures are pystray and dbus_next missing in your environment, not the code.

Keep both things you flagged as going beyond the one-line gate. The field on `WFDPeer` is the correct fix, since `print_scan` was parsing a string formatted for humans, which is the same coupling that caused the bug. And the explicit "not a sink" line closes the second half of #121. The tri-state is right too.

One problem, and it is the mirror image of the case you guarded against. `_nm_get_property` returns an empty string when the gdbus call fails, not only when the property is empty. I confirmed that against a live NetworkManager: a read for a peer that is gone exits 1, so `wfd_ies_raw` is `""`, `wfd_ies_list` is `[]`, and `wfd_capable` becomes `False`. A peer that ages out mid-scan then gets labelled `probably not a Miracast sink`. Before this PR that case printed nothing, and this is the default path.

One line in `nm.py`:

```python
wfd_capable = bool(wfd_ies_list) if wfd_ies_raw else None
```

Empty string means the read failed, `(<@ay []>,)` means it succeeded with no data. I tried it: the failed read goes back to printing nothing and the suite stays green. Worth a test next to the wpa_cli one.

I am testing this against a real TV tomorrow to be sure of the rest. Push that line and I will merge.

Jemo121:
Good catch, and thanks for reverting the source to check the tests actually fail - that is the part I would not have been able to claim for myself.

You are right, and it is worse than the case I did guard against, because it is the default path. I reproduced it before changing anything: a peer whose `WfdIEs` read fails comes back with `wfd_ies_raw = ""`, and a real LG TV prints

```
  [0] [LG] webOS TV  via NetworkManager
      no WFD capability data - probably not a Miracast sink
```

Which is exactly the false negative I wrote the wpa_cli guard for, and I put the same bug back one function over. The distinction I missed is that `""` and `(<@ay []>,)` are different answers: one is a failed read, the other is a successful read of nothing.

Pushed your line, with a comment saying why the two empty-looking values are not the same thing, and a test next to the wpa_cli one. Both directions verified - the test fails when I revert the fix, and the three states now print as:

```
  [0] [LG] webOS TV        WFD capability data detected
  [1] HP LaserJet          no WFD capability data - probably not a Miracast sink
  [2] TV that aged out     (nothing)
```

Kept it as a separate commit so you can diff just the fix against what you reviewed.

Full suite is green here too now - 121 tests, and the only remaining failure is `test_tray_config` from `pystray` missing in my environment. The `dbus_next` ones you saw me report earlier are gone; I installed it while debugging something unrelated, which is a small argument for #129 by itself.

No rush on the TV test, and thanks for doing it.

IlyaP358 (changes requested):
Ran it on my own hardware. Your detection is correct: with the TV in Screen Share mode it reports `WFD capability data detected` and the printer reports the opposite. Yesterday's `None` fix is right too, none of this walks that back.

I am raising the wording because I fell for it myself. I scanned with the TV on a normal input, saw `probably not a Miracast sink` under a Samsung I stream to every day, and spent a while convinced something was broken. Nothing was. A sink only advertises Wi-Fi Display while it is waiting for a connection, so the scan cannot tell "not a sink" apart from "a sink not in screen-share mode yet", and the second is the common case. If it caught me it will catch everyone, except they will not investigate.

So that line should report state, not deliver a verdict. Something like `not advertising Wi-Fi Display; if this is a TV, put it into Screen Share mode first` is accurate and is also the fix most of the time.

The second thing is a real gap. Raw bytes from three devices:

```
Samsung TV, screen share on   00 00 06 01 11 1c 44 00 36   info 0x0111 -> Primary Sink
my other laptop, FluxCast     00 00 06 00 10 1c 44 00 c8   info 0x0010 -> Source
HP LaserJet                   (empty)
```

The laptop bytes are byte-for-byte what `_wfd_ie_device_info` builds, so that is FluxCast advertising itself as a source and `bool(wfd_ies_list)` calls it a sink. Presence is not the question, device type is: bits 0 and 1 of the Device Information bitmap, 0 source, 1 primary sink, 2 secondary sink, 3 both. They sit at `i+3`/`i+4` in the same subelement walk `_parse_wfd_ies_rtsp_port` already does, so a companion parser beside it plus `(info & 3) != 0` covers it. On the wpa_cli path it is the first four hex digits of `wfd_dev_info=`.

The tri-state stays, I am only adding a fourth case for it to distinguish.

Jemo121:
Both pushed.

**The wording.** Took your line nearly as written. It now reads `not advertising Wi-Fi Display; if this is a TV, put it into Screen Share mode first`. You're right about which case is common — I had it backwards, treating the quiet TV as the edge case rather than the default.

**Device type.** `_parse_wfd_ies_device_type` sits beside `_parse_wfd_ies_rtsp_port` as you suggested, and both now walk the stream through one shared `_wfd_device_info_body`, so the two readers can't drift about where subelement 0 starts. The gate is `(info & 3) != WFD_DEVICE_TYPE_SOURCE`.

That's the fourth case, and `WFDPeer` carries `wfd_device_type` so the two Falses stay apart:

| advertised | `wfd_capable` | printed |
|---|---|---|
| sink | `True` | `WFD capability data detected` |
| source | `False` | `advertising Wi-Fi Display as a source, not a sink - another sender, not a display` |
| nothing | `False` | `not advertising Wi-Fi Display; ... Screen Share mode first` |
| read failed | `None` | nothing |

**One judgement call.** A peer that advertises IEs we can't parse — present, but no readable subelement 0 — I made `None` rather than `False`, on the same reasoning as the failed read. Say if you'd rather that be `False`.

**One thing I couldn't settle, so I handled both.** You said the type is the first four hex digits of `wfd_dev_info=`, i.e. the 6-byte Device Information body on its own. Reading wpa_supplicant's source, `p2p_peer` emits `wfd_subelems=` as the whole subelement buffer with headers, and the `P2P-DEVICE-FOUND` event hexes that same buffer into `wfd_dev_info=0x...`. Those two layouts put the bitmap at different offsets and I've no sink here to decide it. So `_wfd_capability_from_hex` reads the value as a subelement stream first and falls back to treating a bare 6-byte value as the body — right either way, and it tolerates the `0x` prefix. If you can paste a real `p2p_peer` block off the Samsung, I'll drop whichever branch is dead.

**Verification.** Your three captures:

```
Samsung TV, screen share on   -> capable=True   device_type=1  port=7236
laptop running FluxCast       -> capable=False  device_type=0  port=7236
HP LaserJet                   -> capable=False  device_type=None
```

The shared-walk refactor is behaviour-identical: I ran the new `_parse_wfd_ies_rtsp_port` against the old implementation over 200k random byte strings, zero mismatches.

Suite is 130 tests, green apart from the `pystray` ImportError in my environment. The source test is pinned against `_wfd_ie_device_info` rather than a byte literal, so if what FluxCast advertises ever changes, FluxCast still can't start looking like a sink to the next FluxCast.

Also rebased onto `dev` after the packaging merge, so the branch is current and the suite is green on top of it.

IlyaP358 (approved):
lgtm! Everything looks great. 
Thank you so much for the thorough work. Merging!

PR #126, opened 2026-09-12
Report ufw port state in --doctor when running unprivileged (#98)

### LINKED ISSUE #98: ufw blocks passive WFD RTSP connection on port 7236
## Summary

When connecting to a sink-only TV with ufw enabled, Wi-Fi Direct activation succeeds but the TV never opens the passive RTSP connection to FluxCast. FluxCast's source-initiated probe to `192.168.49.1:7236` is refused and the session waits indefinitely.

## Logs

```text
[FluxCast WFD RTSP] Server listening on 0.0.0.0:7236
[FluxCast WFD] P2P link is activated; waiting for RTSP session...
[FluxCast WFD RTSP] No passive connection; trying Source-initiated RTSP probe...
[FluxCast WFD RTSP] Active probe: TV=192.168.49.1; connecting to RTSP port 7236...
[FluxCast WFD RTSP] Active probe: TV port 7236 refused — Sink-only device; waiting for its passive connection to us.
```

## Workaround

Allow the WFD RTSP and discovery ports in ufw:

```bash
sudo ufw allow 7236/tcp
sudo ufw allow 5353/udp
```

After adding these rules, retry the connection.

## Expected behavior

The doctor or connection setup should detect/report the ufw requirements, or FluxCast should provide a guided way to allow the WFD ports when ufw is active.

### LINKED ISSUE #141: --doctor misses an enabled UFW and reports "no supported firewall front-end detected"
## Summary

`--doctor` reports a false all-clear (`skip: no supported firewall front-end detected`) on a host where **UFW is enabled and is actively dropping the sink's RTSP connection** to port 7236. The WFD session then never starts and the only log line is the misleading `Active probe: TV port 7236 refused — Sink-only device; waiting for its passive connection to us.`

## Environment

- FluxCast: `0.2.6-0-g58ad629` (branch main), installed from source
- Distro: Arch Linux / omarchy, Hyprland (Wayland)
- NetworkManager: 1.58.1-1, `p2p-dev-wlp2s0`
- Kernel: 7.2.5-3-omarchy
- Wi-Fi: Intel Wireless 8260 (iwlwifi)
- Sink: LG OLED77C15LA (webOS), discovered as `[LG] webOS TV OLED77C15LA`
- Firewall: `ufw 0.36.2-7` + `ufw-docker`, `ENABLED=yes` in `/etc/ufw/ufw.conf`

## Doctor output

```text
skip   firewall  no supported firewall front-end detected (checked ufw and firewall-cmd; open port 7236/tcp manually if a firewall blocks it)
```

## What actually happens

P2P connects, the sink is the GO (`192.168.49.1`), FluxCast is the client (`192.168.49.10`), ping works, and the RTSP server is listening on `0.0.0.0:7236`. The sink then tries to open the RTSP connection, but UFW drops the inbound SYN from the P2P subnet:

```text
[UFW BLOCK] IN=p2p-wlp2s0-0 OUT= ... SRC=192.168.49.1 DST=192.168.49.10 LEN=60 PROTO=TCP SPT=45078 DPT=7236 ... SYN
[UFW BLOCK] IN=p2p-wlp2s0-2 OUT= ... SRC=192.168.49.1 DST=192.168.49.10 LEN=60 PROTO=TCP SPT=45172 DPT=7236 ... SYN
[UFW BLOCK] IN=p2p-wlp2s0-2 OUT= ... SRC=192.168.49.1 DST=192.168.49.10 LEN=373 PROTO=UDP SPT=38573 DPT=52595
```

(The sink's SYN was being dropped the whole time; with `tcpdump`/journal you would otherwise see no traffic at all, which matches #72.)

After allowing the P2P subnet the full handshake works immediately (M1 → M3 → M4 → M5 → SETUP → PLAY, 1080p30):

```text
[FluxCast WFD RTSP] TV connected from 192.168.49.1:45344; local=192.168.49.10
[FluxCast WFD RTSP] <- response for M3_GET_PARAMETER: 200 OK
[FluxCast WFD RTSP] TV RTP port: 53008; source port: 19002; audio=LPCM ..., AAC ...
[FluxCast WFD RTSP] PLAY accepted; media stream started.
```

Workaround used: `sudo ufw allow from 192.168.49.0/24 comment 'FluxCast WFD (P2P subnet)'`

## Root cause

`diagnostics.py::_ufw_check` bails out to `None` whenever `ufw status` is denied:

```python
if result.returncode != 0:
    # ufw status needs root; skip rather than warn when privileges are missing.
    if re.search(r"need to be root|permission denied", output, re.IGNORECASE):
        return None
```

`_firewalld_check` returns `None` too (firewalld not installed), so `_firewall_check` falls through to the `STATUS_SKIP` "no supported firewall front-end detected" result. Because UFW has no unprivileged way to read its status, an **enabled** UFW is treated exactly like **no firewall at all** — the worst possible reading for a diagnostic whose whole job is to catch this (#98).

## Suggested fixes

1. Detect that UFW is *enabled* without root and stop reporting `skip`:
   - `systemctl is-active ufw` (unprivileged), and/or
   - parse `/etc/ufw/ufw.conf` for `ENABLED=yes`.
   When enabled but unverifiable, emit at least a `warn`: "ufw is enabled but its rules need root to read; if the sink never connects, run `sudo ufw allow from 192.168.49.0/24`".
2. Consider reading `/etc/ufw/user.rules` (world-readable) to check for an existing allow rule covering 7236/tcp or the P2P subnet before warning.
3. The manual hint should point at the **P2P subnet**, not only `7236/tcp`: RTSP (7236/tcp), the RTCP/RTP back-channel and the sink's UDP traffic all ride the same link, so `sudo ufw allow from 192.168.49.0/24` is the robust one-liner.
4. `wfd/firewall.py::_open_wfd_firewall_port` only handles firewalld, so on UFW hosts FluxCast can never open the port itself; a UFW path (or at least the correct hint) would help. Note some hosts can have BOTH front-ends installed, so the fix should not assume firewalld-only.

Happy to test a patch on the LG OLED77C15LA setup. Thanks for the tool!

Jemo121:
## Summary

Fixes #98. The cause turned out to be different from what I assumed when I picked this up, so worth stating plainly: `--doctor` has checked whether ufw allows `7236/tcp` since #50. The check just never reached the person who needed it.

`ufw status` refuses to run as a normal user, and `_ufw_check` returned `None` on that branch - skipping the check rather than warning. `--doctor` is normally run unprivileged, so on a ufw host the one line that would have named port 7236 left the report entirely. That matches #98 exactly: nothing in the diagnostics, then a session that waits for an RTSP connection the firewall is dropping.

## Fix

Fall back to the `ENABLED` line in `/etc/ufw/ufw.conf`, which ships world-readable, so an unprivileged `--doctor` can still tell whether ufw is on:

| ufw state | Reported as |
|---|---|
| disabled | `ok` - installed but disabled, port not blocked |
| enabled | `warn` - names the port and `sudo ufw allow 7236/tcp` |
| unreadable | `warn` - could not verify |

Deliberately not `systemctl is-active ufw`. On this machine the unit reports `active` while `ufw.conf` says `ENABLED=no`, so that signal would warn on hosts with no firewall actually running.

Before, on a ufw host without root, the report had no ufw line at all. After:

```
  warn   firewall (ufw)         ufw is enabled; port 7236 needs root to verify
                                open it with: sudo ufw allow 7236/tcp
```

## Also on the connect path

Covered in a comment below - it goes a little past the doctor-only reading of the issue, and is easy to drop.

## Testing

Ran the real `--doctor` on a ufw host unprivileged: the line now appears (`ufw is installed but disabled`) where before there was nothing.

8 new tests covering all three states, the conf parsing, an unreadable conf, and the connect-path hint. Suite goes 112 -> 120; the 6 failures are the same ones pristine `dev` has here - `pystray` and `dbus_next` are not installed in my environment.

### DIFF
--- a/src/diagnostics.py
+++ b/src/diagnostics.py
@@ -630,6 +630,22 @@ def _supplicant_wfd_check() -> Check:
     )
 
 
+def _ufw_enabled() -> Optional[bool]:
+    """Whether ufw is switched on, without needing root. None if unreadable.
+
+    `ufw status` refuses to run as a normal user, so read the ENABLED line
+    from ufw.conf, which ships world-readable. Not `systemctl is-active ufw`:
+    the unit can be active while ufw itself is disabled.
+    """
+    try:
+        with open("/etc/ufw/ufw.conf") as conf:
+            match = re.search(r"^\s*ENABLED\s*=\s*(\w+)", conf.read(),
+                              re.MULTILINE | re.IGNORECASE)
+    except OSError:
+        return None
+    return match.group(1).lower() == "yes" if match else None
+
+
 def _ufw_check() -> Optional[Check]:
     if not shutil.which("ufw"):
         return None
@@ -640,10 +656,27 @@ def _ufw_check() -> Optional[Check]:
         return Check("firewall (ufw)", STATUS_WARN, "could not query ufw status", str(exc))
 
     output = (result.stdout + result.stderr).strip()
+    # ufw prints the root error and still exits 0, so match the text rather
+    # than trusting the return code.
+    if re.search(r"need to be root|permission denied", output, re.IGNORECASE):
+        # Returning None here is what let #98 happen: --doctor is normally run
+        # unprivileged, so the one line that would have named port 7236 left
+        # the report entirely and the session just hung with no explanation.
+        enabled = _ufw_enabled()
+        if enabled is False:
+            return Check("firewall (ufw)", STATUS_OK,
+                         "ufw is installed but disabled; port not blocked",
+                         "read from /etc/ufw/ufw.conf; ufw status needs root")
+        detail = (f"open it with: sudo ufw allow {WFD_RTSP_PORT}/tcp\n"
+                  "  or check the current rules with: sudo ufw status")
+        if enabled is None:
+            return Check("firewall (ufw)", STATUS_WARN,
+                         f"could not verify whether ufw allows port {WFD_RTSP_PORT}",
+                         detail)
+        return Check("firewall (ufw)", STATUS_WARN,
+                     f"ufw is enabled; port {WFD_RTSP_PORT} needs root to verify",
+                     detail)
     if result.returncode != 0:
-        # ufw status needs root; skip rather than warn when privileges are missing.
-        if re.search(r"need to be root|permission denied", output, re.IGNORECASE):
-            return None
         return Check("firewall (ufw)", STATUS_WARN, "ufw status query failed", output)
 
     if re.search(r"Status:\s*inactive", output, re.IGNORECASE):
--- a/src/wfd/firewall.py
+++ b/src/wfd/firewall.py
@@ -1,6 +1,6 @@
 import subprocess
 
-from diagnostics import _firewalld_active
+from diagnostics import _firewalld_active, _ufw_enabled
 
 from .proc import _run
 
@@ -22,6 +22,26 @@ def _print_firewall_manual_hint(port: int, reason: str) -> None:
         "  or pass --wfd-no-firewall if you manage the firewall yourself."
     )
 
+def _warn_if_ufw_may_block(port: int) -> None:
+    """Say something before the session hangs on a ufw host.
+
+    FluxCast cannot open the port itself here: ufw has no Polkit integration
+    to prompt through the way firewalld does, so this would mean shelling out
+    to sudo mid-session to edit the user's firewall. Printing the command is
+    the honest limit (#98).
+
+    Deliberately file-only - no subprocess. The firewall probe runs on the
+    connect path, and a blocking call here was already a problem once (#114).
+    """
+    if _ufw_enabled() is not True:
+        return
+    print(
+        f"[FluxCast WFD] ufw is enabled. If the sink never opens its RTSP "
+        f"connection, allow port {port}/tcp:\n"
+        f"    sudo ufw allow {port}/tcp\n"
+        "  or pass --wfd-no-firewall to silence this."
+    )
+
 def _open_wfd_firewall_port(port: int) -> bool:
     """
     Runtime-only (no ``--permanent``): cleared on reload/reboot, and removed on
@@ -30,6 +50,7 @@ def _open_wfd_firewall_port(port: int) -> bool:
     """
     # None ("couldn't ask systemd") is treated like inactive: nothing to open.
     if not _firewalld_active():
+        _warn_if_ufw_may_block(port)
         return False
 
     # Polkit can gate the read-only query as well, so it gets the same budget
--- a/tests/test_diagnostics.py
+++ b/tests/test_diagnostics.py
@@ -172,12 +172,46 @@ class FirewallCheckTest(unittest.TestCase):
             check = diagnostics._firewall_check()
         self.assertEqual(check.status, diagnostics.STATUS_WARN)
 
-    def test_ufw_without_root_is_skipped(self):
-        err = "ERROR: You need to be root to run this script"
+    # `ufw status` refuses to run unprivileged, and --doctor is normally run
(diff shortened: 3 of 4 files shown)


Jemo121:
The one thing here that goes past a strict reading of #98, easy to cut if you would rather keep this to `--doctor`.

**The connect path now prints the port and command too.** #98 asks for "the doctor **or** connection setup" to report this, and the connection setup half was silent: on a ufw host `_open_wfd_firewall_port` finds no firewalld, returns `False`, and prints nothing. So someone who never ran `--doctor` still gets the original symptom - a session that sits waiting for an RTSP connection with no explanation.

```
[FluxCast WFD] ufw is enabled. If the sink never opens its RTSP connection, allow port 7236/tcp:
    sudo ufw allow 7236/tcp
  or pass --wfd-no-firewall to silence this.
```

Two deliberate limits on it:

**It does not open the port.** Mirroring the firewalld path properly is not possible - firewalld has Polkit, so `firewall-cmd --add-port` can prompt and the session opens the port for its duration and closes it on exit. ufw has no equivalent, so the same behaviour would mean shelling out to `sudo` mid-session to edit the user's firewall, and it could not be reliably undone afterwards. Printing the command seemed like the honest limit, but if you would rather FluxCast attempt it behind a flag, say so and I will add it.

**It runs no subprocess** - it reads `ufw.conf` and nothing else. This sits on the connect path, and #114 was a blocking firewall probe there causing a Polkit dialog and a 3s stall, so I did not want to reintroduce that shape. There is a test pinning the no-subprocess property.

It is also gated by `--wfd-no-firewall` already, since it lives inside the call `session.py` skips under that flag.

IlyaP358:
Keeping both halves. The doctor fix is what the issue asks for, but the session hint is the part that reaches the person who filed #98, since they never ran `--doctor` at all, they just got a session that hung. That was the actual bug.

No to the flag that opens the port. Shelling out to sudo mid-session to edit someone's firewall, with no reliable way to undo it, is worse than printing the command. Your reasoning for not doing it was right.

One thing to fix. `session.py` calls `_open_wfd_firewall_port` twice when `--wfd-uibc` is on, so the hint prints twice and the second copy says "if the sink never opens its RTSP connection, allow port 7239/tcp", which is not what 7239 is. Print it once per session.

Nothing else, the rest holds up. Push that and I will merge it =]

Jemo121:
Fixed. You're right that the second copy was worse than noise — it named 7239 as the port RTSP arrives on.

Moved the hint out of `_open_wfd_firewall_port` and into one call from `session.py` that takes every port the session listens on, each line saying what its port is:

```
[FluxCast WFD] ufw is enabled. If the sink never opens its RTSP connection, allow the ports FluxCast listens on:
    sudo ufw allow 7236/tcp    # RTSP
    sudo ufw allow 7239/tcp    # UIBC
  or pass --wfd-no-firewall to silence this.
```

Without `--wfd-uibc` it's the single RTSP line, as before.

I moved the call rather than adding an "already printed" flag, deliberately. A module-level flag would go wrong in the tray applet, which starts and stops sessions inside one long-lived process — the hint would appear on the first session and then never again. Moving it makes "once per session" structural rather than something to remember to reset.

One consequence worth flagging: the hint now carries its own `_firewalld_active()` guard. That used to be implicit, since the only caller reached it in the not-firewalld branch — now that `session.py` calls it unconditionally, a firewalld host would otherwise get a ufw hint it doesn't need. `test_firewalld_host_stays_quiet` covers that.

Four tests added, including one asserting a single hint with both ports labelled, and one pinning `_open_wfd_firewall_port` to printing nothing of its own. Suite is 144 tests, green apart from the `pystray` ImportError in my environment.

IlyaP358 (approved):
nice, all good, merging. The labelled ports read much better than the old copy, and moving the call was the right call over a flag.

PR #132, opened 2026-09-12
feat: add Cosmyx IWD and COSMIC portal support

ThiagoHaeitmann:
This PR adds support for NetworkManager using IWD as the Wi-Fi backend and adds detection for xdg-desktop-portal-cosmic.

The main changes are:

Detect Wi-Fi P2P devices using NetworkManager DeviceType instead of relying on "p2p" being present in the interface name.
Support IWD P2P devices exposed as paths such as /net/connman/iwd/0.
Avoid wpa_supplicant-specific device-name and GO-intent operations when IWD is in use.
Increase the NetworkManager P2P activation timeout for IWD.
Add xdg-desktop-portal-cosmic to the recognized portal backends.

Tested on NixOS with:

COSMIC / Wayland
NetworkManager 1.56
IWD backend
xdg-desktop-portal-cosmic

In this environment FluxCast successfully detects the NetworkManager Wi-Fi P2P device and starts WFD scanning through /net/connman/iwd/0.

I made these corrections in my fork and tested them successfully in my environment. The changes included in this PR are the modifications that worked for me and are intended to keep the existing wpa_supplicant behavior unchanged while adding IWD compatibility.


Jemo121:
Hi @ThiagoHaeitmann, welcome! Not the maintainer here, just another contributor - @IlyaP358 has the final say on all of this. But you did the legwork on a setup nobody else here runs, so I wanted to give you feedback sooner rather than let it sit.

For what it is worth, this is a good contribution. You found a real assumption in the code, explained it clearly, tested it on actual hardware, and kept the diff to 56 lines across three files. Using `DeviceType` instead of sniffing the interface name for `"p2p"` is the correct fix, not a workaround - the name test was always a guess and you replaced it with the thing NetworkManager actually tells us.

One process note: the project merges into `dev` rather than `main` ([CONTRIBUTING.md](https://github.com/IlyaP358/fluxcast/blob/dev/CONTRIBUTING.md)), so this wants retargeting. Nothing else needs redoing - `dev` is ahead of `main` but not in these three files, so the diff applies as-is.

Now the technical part. One thing I think is worth changing, and some smaller notes.

### The `DeviceType` check has no fallback

This is the one I would flag as worth fixing before it goes in, and I say that having made the exact same mistake last week - @IlyaP358 caught it in #125, so it is fresh.

`_nm_get_property` returns `""` when the gdbus call *fails*, not only when the property is genuinely empty. So:

```python
device_type = _variant_uint(_nm_get_property(path, ..., "DeviceType"))
if device_type != 30:
    continue
```

If that read fails for any reason, `_variant_uint("")` returns `None`, `None != 30` is `True`, and the device is skipped. With the old code the name test would still have found it; now nothing does, and the user gets `NetworkManager P2P device disappeared before connection` with no clue why.

I checked this rather than guessing:

```
_variant_uint('') -> None
so `device_type != 30` is: True -> the device is SKIPPED
```

Keeping the old test as a fallback covers it, and costs one branch:

```python
if device_type is not None:
    if device_type != 30:
        continue
elif not iface or "p2p" not in iface.lower():
    # DeviceType unreadable - fall back to the old name test rather than
    # discarding a device we cannot classify.
    continue
```

That way IWD works via `DeviceType`, and a failed read degrades to today's behaviour instead of finding nothing.

### Smaller things, all optional

**The device path is now looked up before the scan.** You moved `_nm_p2p_device_path` above `_scan_and_select`, which you had to, since `using_iwd` gates `_set_p2p_device_name`. The old ordering looked deliberate though: the path was fetched *after* the scan, and the error says "disappeared **before connection**", which reads like a freshness check for a device vanishing during those 8-30 seconds. The path fetched early is then used for `_connect_peer` later.

@IlyaP358 - was the scan-then-lookup order deliberate, and does it matter if the path is now read before the scan? Re-fetching after the scan would keep both properties at the cost of one extra call, but that is your call, not something Thiago should guess at.

**`--wfd-interface` is effectively ignored under IWD.** Reading this:

```python
if requested and requested not in (iface or ""):
    if not (iface or "").startswith("/net/connman/iwd/"):
        continue
```

any `/net/connman/iwd/*` device matches regardless of what the user asked for.

@IlyaP358 - is that acceptable? There is normally only one P2P device so it likely never bites, but it does mean `--wfd-interface` is silently ignored on IWD setups, and I did not want to ask for a change to flag behaviour on your behalf.

**`30` and `150.0`/`35.0` are inline literals.** `dbus.py` already keeps NM enums as module constants (`NM_DEVICE_STATE_NAMES` and friends), so a named `NM_DEVICE_TYPE_WIFI_P2P = 30` would sit naturally there. And `timeout=150.0 if using_iwd else 35.0` restates the function's own default, so the two can drift.

@IlyaP358 - do you want those pulled out as constants, or is inline fine here? Style call rather than a correctness one.

**Tests.** `_nm_p2p_device_path` is pure logic over mocked gdbus output, so it tests easily - `tests/test_peers.py` on `dev` has close examples if you want a pattern.

@IlyaP358 - do you want tests on this before it goes in? It would pin the IWD path so nobody breaks it later without the hardware to notice, but you may not want to ask that of a first contribution.

None of that takes away from the main point: the diagnosis is right and the fix is the right shape. Retarget to `dev` and add the fallback, and I think this is in good shape for @IlyaP358 to take a look. Everything under "smaller things" is his call rather than mine, so I would wait for his answer before changing any of it. Happy to help if anything above is unclear.

IlyaP358:
Thanks both. @Jemo121, that review is right, including the fallback, and thanks for leaving the judgement calls to me instead of deciding them.

@ThiagoHaeitmann, welcome, and this is a good first contribution! `DeviceType` instead of matching `"p2p"` in the interface name is the correct fix and I want it regardless of IWD, since the name test was always a guess.

I was going to ask whether you actually got a picture on screen, but you answered that in #133 before I asked, and better than I would have: discovery and activation work, the connection dies inside IWD's GO negotiation and then its WSC handling, and you separated the two layers yourself. That is the right split, and it tells me how to describe this when it lands. FluxCast will detect and scan on NetworkManager with IWD, and streaming does not work yet for reasons outside this repo. I would rather say that plainly than have someone install it expecting a working stream.

One consequence for this PR: since the 150 second timeout comes from a negotiation that fails anyway, drop it and keep the default until there is an IWD setup where the longer wait actually completes. Right now it only stretches the failure from thirty seconds to two and a half minutes.

Answers to @Jemo121's four questions:

The lookup/scan order was not deliberate, it came from a module split. Keep the lookup after the scan anyway: `_scan_and_select` retries up to three times, so the gap before connecting can be around 30 seconds, and `_nm_scan` looks the path up itself regardless, so the extra call costs nothing.

`--wfd-interface` silently ignored under IWD is not acceptable. Either match the requested string against the path too, or print one line saying the flag does not apply there.

Yes to the constant, `NM_DEVICE_TYPE_WIFI_P2P = 30` in `dbus.py` with the other NM enums. For the timeout, see above, it should just go.

Yes to tests, small ones. `_nm_p2p_device_path` over mocked gdbus: a normal device, an IWD path, and a failed `DeviceType` read falling back to the name test. That last case is the part nobody else can check without your hardware.

I'm also retarget it to `dev`, as @Jemo121 mentioned.

IlyaP358 (approved):
lgtm, merging =] 
Thanks for turning this around quickly, and for the tests especially, they pin the IWD path so nobody can break it later without owning your hardware.

I'll add you on [contributors page](https://fluxcast.dev/contributors.html).

Issue #128, opened 2026-09-12
Opt-in VAAPI/QSV encode, sink mode-state file, and M16 Session keepalive hardening`


realchrisolin:
## Summary

Proposal for an opt-in hardware encode path on the wlroots/wf-recorder capture pipeline, plus a couple of related WFD helpers, without changing default behavior for existing FluxCast users.

Branch: https://github.com/realchrisolin/fluxcast/tree/feat/wfd-hw-encode-modes

## Motivation

On Hyprland laptops with Intel VAAPI, software libx264 works but burns more CPU/power than necessary for Miracast. I also needed:

1. A way for an external UI to read the sink’s advertised CEA/VESA modes after RTSP negotiation.
2. More reliable M16 keepalives on a portable Miracast sink that returns **454** if `Session` includes `;timeout=30`, then TEARDOWNs when keepalives fail.

## Design (non-breaking defaults)

| Behavior | Default (unchanged) | Opt-in |
|----------|---------------------|--------|
| Video encoder | `libx264` | `FLUXCAST_WFD_ENCODER=va
| `wf-recorder -D` | kept (continuous capture) | omit
| Battery / power-saver bitrate bias | off for default
| Mode-state JSON | off | `FLUXCAST_WFD_MODE_STATE=/pa

Also:

- `--doctor` gains a **warn** when `h264_vaapi` / `h264_qsv` is available but encoder still defaults to libx264 (`set FLUXCAST_WFD_ENCODER=auto …`).
- M16 keepalives use a **bare** `Session` id and reschedule at **20s**; capture rebind (`SIGUSR1`) does not kill the keepalive/probe chain.
- Env vars documented under **WFD environment variables** in `documentation/DOCUMENTATION.md`.

## Tested

- **OS / compositor:** Arch-based Omarchy, Hyprland
- **Wi-Fi:** Intel (AX-class), NetworkManager P2P
- **Sink:** HotYeah-class portable Miracast monitor (WFD 1.0, AOSP-style TS with `--wfd-aosp-pmt-pid`)
- **Result:** `FLUXCAST_WFD_ENCODER=auto` → `h264_vaapi on /dev/dri/renderD128`; PLAY accepted; keepalives `200 OK`; stable streaming

Unit coverage added for encode defaults, Device-scoped battery ignore, mode-state filtering, `-D` vs damage env, bare-Session keepalives, and the doctor hint.

## Thoughts

[...]

Issue #41, opened 2026-06-09
ADD LPCM audio in MPEG-TS support required by Microsoft 4k Wireless Adapter


IlyaP358:
Standard GStreamer `mpegtsmux` pipelines fail to maintain a stable connection with the Microsoft Wireless Display Adapter when streaming LPCM audio. This happens because the Microsoft adapter strictly requires a specific LPCM packet format (private stream type `0x83` combined with a custom `0xA0 0x06 0x00 0x09` header extension) and is extremely sensitive to PCR (Program Clock Reference) timing jitter.

Solution is implementing a custom, lightweight Python-based MPEG-TS/RTP multiplexer layerin the `lpcm-audio` branch.

IlyaP358:
@SebSch182 the lpcm-audio branch is ready for testing.
run: 
`git fetch origin`
`git checkout lpcm-audio`
`git pull origin lpcm-audio`

If you have some time to test it out with your adapter, let me know how it goes!

SebSch182:
It works! But the sound is terrible. Like it's coming through a tin can.

IlyaP358:
Nice, could you test the latest commit on **lpcm-audio**? I switched the audio byte order from big-endian to little-endian (S16LE). If the "tin can" effect disappears and audio sounds normal, that was the issue. If it's still bad, let me know and i'll dig deeper.

SebSch182:
It’s even worse. Only hissing and rushing sounds.

IlyaP358:
Unfortunately, little-Endian turned out to be wrong =[ , so I reverted back to Big-Endian.

Maybe i found a issue with the headers and byte alignment. I've just pushed a new fix that adds a proper DVD LPCM audio frame header and aligns the stream pointers perfectly according to the spec.

SebSch182:
Still ugly audio.

IlyaP358:
Thanks for testing. 
Since I don't own this Microsoft adapter myself, i'm basically debugging a black box. Your feedback is the only thing helping me cut through the guesswork here.

To help me rule out a bunch of wrong hypotheses and figure out exactly what's wrong with the audio headers, could you describe what "ugly" means in a bit more detail? It would help a ton if you could check these points:

1. Can you actually tell what's playing? Is it recognizable as speech or a song but heavily distorted, or is it just completely broken, scrambled noise?

[...]

Issue #19, opened 2026-05-16
Optionally enable hardware video encoding


alba4k:
This could be done by allowing to specify extra arguments for `wf-recorder`. Normally, hw video encoding is enabled via `wf-recorder -c h264_vaapi -d /dev/dri/renderD128` (as per [their readme](https://github.com/ammen99/wf-recorder))
The benefits could be lower latency, cpu usage and lower power usage. Assuming this can work with the rest of the pipeline. Platforms that use `xdg-video-portal` for screen capture should be already accelerated by default (can xdph not be used on hyprland over wf-recorder?)

alba4k:
Ok I tested this a bit and the latency seems to be quite a lot higher when using this (like, 1-2 seconds higher), likely due to `h264_vaapi` being slower than `rawvideo`. I think a faster encoding might actually end up outperforming the uncompressed stream in usecases where the network speed is the bottleneck

I did notice cpu usage from `wf-recorder` going down from 40-50% to around 25%, which would probably also result in considerable power saving.

I can open a pr if you think this is worth investigating further (maybe handled by a `--enable-wf-recorder-vaapi` flag). 

If I were to open a pr, may I ask what the `wf-recorder` command in `capture.py` is used for? should that one be touched?

IlyaP358:
Hello @alba4k, thanks for testing this and sharing the results!

On capture.py that file is only used for the DLNA and Chromecast protocols (HLS streaming). For WFD, the wf-recorder command lives in wfd.py in `_start_desktop_wf_recorder()` there wf-recorder outputs rawvideo to stdout, and ffmpeg reads it, encodes with libx264 (-tune zerolatency) and sends it over RTP. You'd only need to touch `wfd.py`.

On vaapi: the +1–2s latency you measured is unfortunately a dealbreaker for WFD as a default, latency is the primary metric for Miracast. That said, as an opt-in flag (e.g. --wfd-wf-recorder-vaapi) for users who consciously want CPU/power savings at the cost of latency, it makes sense. A PR along those lines is welcome =]

also regarding wf-recorder on Hyprland. xdg-desktop-portal-hyprland is already supported. You can use it right now with `--wfd-capture-backend portal` on Hyprland but it is quite unstable and sometimes during tests I have breaks, why so far I have not been able to figure out and focused specifically on supporting wf-recorder. 

wf-recorder is just the default there because it's more straightforward and reliable on wlroots-based compositors.

alba4k:
[...]

PR #135, opened 2026-09-13
feat(wfd): opt-in VAAPI/DMA-BUF, LPCM mux, and ICC wf-recorder

realchrisolin:
## Summary

Opt-in improvements for wlroots/`wf-recorder` Miracast on Hyprland without changing defaults for existing FluxCast users:

1. **Hardware encode / RENDER ENGINE** — VAAPI (and QSV) encode with DMA-BUF CQP preference, GPU→CPU fallback, live preference file for external UIs.
2. **LPCM for LPCM-only sinks** — WFD LPCM mux (`stream_type=0x83`, AOSP-style PIDs) when the sink does not take AAC; Pulse capture via **`pw-cat --target`** of a dedicated null-sink monitor (not mic / not Lavf `ffmpeg -f pulse`).
3. **ICC `wf-recorder` opt-in** — `FLUXCAST_WFD_WF_RECORDER_BIN` + `PROTO=icc`; capture `-r` follows stream `config.fps`; damage-aware via `FLUXCAST_WFD_WF_RECORDER_DAMAGE`.
4. **Capture rebind** — `SIGUSR1` rebuilds desktop capture without dropping RTSP; geometry-drift / stagnant-TX / dead-sender recovery.
5. **Mode state + M16** — optional sink mode JSON for UIs; keepalive hardening for sinks that reject `Session;timeout=…`.

Closes / continues: https://github.com/IlyaP358/fluxcast/issues/128

## Defaults (non-breaking)

| Knob | Default (unchanged) | Opt-in |
|------|---------------------|--------|
| Encoder | `libx264` | `FLUXCAST_WFD_ENCODER=vaapi\|qsv\|auto` |
| Capture encode path | raw pipe | `FLUXCAST_WFD_CAPTURE_ENCODE_PREF=dmabuf` (or `vaapi` / `cpu`) |
| `wf-recorder -D` | continuous (`-D`) | `FLUXCAST_WFD_WF_RECORDER_DAMAGE=1` omits `-D` (LPCM + DMA paths) |
| ICC binary | PATH stock | `FLUXCAST_WFD_WF_RECORDER_BIN=…` + optional `PROTO=icc` |
| Mode-state file | off | `FLUXCAST_WFD_MODE_STATE=/path/to.json` |

## LPCM / audio

- Negotiate LPCM when the sink is LPCM-only; `FLUXCAST_WFD_FORCE_AAC=1` keeps AAC for debugging.
- Capture **`miracast.monitor`-style** Pulse monitor with `pw-cat --target` (unique app name) so Pulse stream-restore cannot remap Lavf onto Speakers.monitor (silent TV / dual-play class of bugs).
- Muxer: AOSP-style framing, video PID `0x1011`, audio `0x1100`, small leaky queues, TS stuffing.

## ICC / capture

- Opt-in ICC client (`ext-image-copy-capture`); not auto-probed.
- ICC `-r` matches negotiated/UI stream FPS (`config.fps`); override with `FLUXCAST_WFD_ICC_CAPTURE_FPS`.
- DMA-BUF + CQP (`FLUXCAST_WFD_VAAPI_QP`, default 18) for desktop sharpness on Intel.

## Tests

```bash
python3 -m unittest \
  tests.test_hw_encode \
  tests.test_mode_state \
  tests.test_capture_restart \
  tests.test_wfd_lpcm_mux \
  tests.test_lpcm_audio_capture \
  tests.test_icc_integration \
  tests.test_wf_recorder \
  tests.test_diagnostics \
  tests.test_wfd
```

(177 tests OK on this branch tip against the suites above. Full
`unittest discover` still hits pre-existing portal/gi failures also present
on upstream `dev` in this environment.)

## Test plan (hardware)

Per CONTRIBUTING — tested on **Hyprland** + portable Miracast sink (LPCM-only TV):

- [x] Picture + LPCM audio on Extend with ICC `wf-recorder` + DMA-BUF VAAPI
- [x] Default output Miracast null sink; capture pinned to `.monitor` (no Speakers.monitor remap)
- [x] Stream mode 1080p30 → wf-recorder `-r 30` / `framerate=30`
- [x] `DAMAGE=1` → LPCM cmdline omits `-D` after reconnect
- [x] `SIGUSR1` / `restart-capture` rebuilds capture without tearing down RTSP
- [ ] Non-Samsung / non-Hyprland (help wanted)

Session latency log / doctor output available on request.

## Notes for reviewers

- Rebased onto current `dev` (includes Debian packaging #122).
- **Relation to merged LPCM (#46 / historical `lpcm-audio` branch):** LPCM mux
  already landed on `dev` via [#46](https://github.com/IlyaP358/fluxcast/pull/46)
  (Microsoft Wireless Display Adapter framing; video PID `0x1000` = PCR).
  This PR **builds on that path** rather than introducing LPCM from scratch:
  AOSP-style PIDs (H.264 on `0x1011`, dedicated PCR `0x1000`) for LPCM-only
  TVs that stalled on “obtaining stream”, `pw-cat` monitor capture, AAC
  fallback / force flags, queue/stuffing harden, plus the separate HW-encode
  and ICC opt-in work. The remote `lpcm-audio` branch tip is already an
  ancestor of `dev` — no competing unmerged LPCM PR.
- Large diff is mostly `hw_encode.py`, `wlroots.py` capture paths,
  `wfd_lpcm_mux.py`, and tests.

## Docs

- `documentation/DOCUMENTATION.md` — `FLUXCAST_WFD_*` env table and LPCM/ICC notes  
- `README.md` — brief pointers where relevant


IlyaP358:
Hi @realchrisolin, thank you for five days of real work, and sorry it took me this long to give you a clear answer. You deserved one sooner.

I am not merging this in its current form. That is about the shape of the change, not about you, so here are the actual reasons.

**It cannot be reviewed.** 73 commits, 48 files, fifteen force-pushes in five days. Whatever Jemo or I read on Monday is a different branch by Tuesday. That is not a complaint about pace, it means no honest review is possible against a moving target.

**It is not opt-in, despite the title.** The scan retry budget goes from 3 to 6 for everyone with no flag. `_default_wifi_interface` is rewritten, and in that default path it now runs an external script:

```python
# Optional external resolver (omarchy-miracast list_p2p_radios.py).
helper = (os.environ.get("FLUXCAST_LIST_P2P_RADIOS") or "").strip()
```

I understand why you need that downstream, but FluxCast cannot shell out to a plugin from its own core. Seventeen of the changed files are outside encoding, LPCM and ICC entirely, including both RTSP files, session, probe, modes and the D-Bus policy. That policy file exists because of a security report with advisories still in progress, widening those grants may be right, but not inside a diff titled after an encoder.

**We cannot test it,** and this is the part I most want to be straight about, because no amount of better code fixes it. Neither Jemo nor I own a VAAPI Miracast setup, a USB P2P dongle, or an LPCM-only sink. Merging nine thousand lines we cannot exercise means shipping on faith to people who use this on their TVs every evening. That is the one promise this project makes, and I will not spend it. For scale, this adds about sixty percent of the existing codebase in one go.

**What I do want.** Both of your headline features are already open goals here, filed by us before you arrived: #19 for hardware encoding and #41 for LPCM, which I opened myself. So this is a route, not a door.

One feature per PR, small enough to read in an evening, defaults unchanged unless the title says otherwise, and each one confirmed on real hardware by whoever owns it. That last rule is not aimed at you: nothing has ever landed in our WFD path without someone holding the device, from the UIBC work on tablets to the LG debugging.

Start with `--wfd-encoder vaapi` behind a flag. No DMA-BUF, no ICC, no P2P, no policy changes. A few hundred lines I can read properly and you can verify on your own hardware. If that goes well, LPCM next, and you are the right person for it since you own the sink.

I am leaving this open as a reference rather than closing it. Convert it to a draft or close it yourself once the first slice is up.

@Jemo121, you self-assigned this and have been carrying it alone. You don't need to burn your weekend on this. This was mine to call and I should have called it three days ago instead of leaving you in a diff that gets rewritten nightly.

Jemo121:
@realchrisolin — I went through this properly before Ilya's call, and the vaapi-first route is the one I'd have landed on too, so I won't re-argue any of it. Here are the concrete things I found that carry forward into the first slice.

**The test suite is what I'd fix ahead of any code.** At current HEAD it's 18 failing out of 334, excluding a `pystray` import error that's my environment. When I went through it on the 15th it was 2. Two clusters account for twelve:

```
6 ×  TypeError: fileno() returned a non-integer
6 ×  AttributeError: 'Harness' object has no attribute 'tx_interface'
```

Both look like harness drift rather than product bugs, but they bury one that isn't. `test_cbr_sets_b_before_rc_mode` expects `b=28000000` and gets `b=12000000`. I bisected it: it passed when it was added in the RC-presets commit, and first fails at `b082bd6 fix: wf-recorder QVBR uses target bitrate + maxrate`. Worth settling whether the test or the change is the wrong one, since that answer probably belongs in the vaapi slice itself.

**One default does change**, and it's easy to miss because the description's table lists defaults as unchanged. `045f55f` takes the VBV multiplier from `2.0` to `0.5` for every non-LG peer, and removes the comment warning that tighter values cause VBV underflow with ultrafast at high resolutions. The sinks that affects are Samsung — #1, #2 and #3 here. If it's wanted, it probably deserves its own PR and its own argument.

**A suggestion I think is free.** `63212f2`, finding the sink IP when the group MAC differs from the discovery MAC, is independently valuable and has nothing to do with encoding. I checked both files it touches: they apply cleanly to `dev` as-is, and `addressing.py` isn't touched by any other commit on the branch, so

```
git cherry-pick 63212f2
```

onto a fresh branch gives a reviewable PR immediately. There's an LG user in #136 right now who may be hitting exactly this. That might be a better first slice than the encoder — smaller, a fix rather than a feature, and it helps someone this week.

**On size.** `src/wfd/media/wlroots.py` goes from 133 to 1308 lines here, which would make it the largest file in the tree by some way. The branch also adds four new files over 300 lines, `usb_dedicated.py` at 728 among them.

Last thing, practical rather than critical: when the vaapi slice is up, I'll review it the same week if it stays put. What made this branch hard to get through wasn't really its size — it was that it moved every day I was reading it.

realchrisolin:
Sorry guys, I should've been better about communication. I jumped the gun on opening this PR, realizing there was a lot more complexity at play than I initially realized and started trying to compensate for it after the fact (and the subsequent changes I made are not all complete or introduce breaking changes, as you've noticed) by committing fixes retroactively. This branch is effectively "what I'm running locally" as I'm trying to get things working. You can use it to get some idea of what I'm working on and speak up if you have any thoughts about the architecture / design.

I think we're all on the same page that all these changes shouldn't be a single PR or committed as-is. I'm testing with a variety different Miracast devices / wireless adapters / antennas / bandwidth / link rates / encoding settings. Let me work on getting things working more stable in re low TX wireless adapters and once I'm satisfied, open separate PRs for #19 and #41 and at least get those improvements in. 

As far as Miracast support goes, I'll reflect on your and Jemo's feedback and see what I can do to get it working without so many changes. Or maybe bring in pieces at a time. If and when the time comes for maintainer testing with a Miracast sink, I'm not opposed to buying one of you one of those portable Miracast displays.

IlyaP358:
@realchrisolin, no worries at all, and thank you for the honest update. Testing across that hardware matrix is genuinely hard, and I understand completely how a local branch turns into that. We are on the same page.

One correction before you spend an evening on it, because things moved since @Jemo121 wrote that. I merged a receiver-authentication change today that rewrote `addressing.py` and `probe.py`, and `63212f2` no longer cherry-picks cleanly onto `dev`, it conflicts in both files. Part of it may also be redundant now: the new code accepts a single unambiguous neighbour on the session's group interface, which covers the case where a sink uses a different address in the group than it advertised during discovery. That is the LG behaviour from #44, not #136, which is a different problem on a ScreenBeam.

So please pull `dev` first and see what is actually left. There may still be something worth having in there, the retry around the empty ARP table in particular, but it is a smaller and different change than it was yesterday.

On the Miracast hardware, thank you, that is a generous offer and I do not take it lightly. Let's revisit it once the first slices are in and we know what specifically needs verifying. I want to be clear that it makes no difference to how your PRs get reviewed either way, they get the same read regardless.

Looking forward to the first ones. And for what it is worth, @Jemo121's read of the branch was the useful kind: he went through it properly rather than judging it by size.

PR #143, opened 2026-09-20
Authenticate WFD receivers and enforce a single RTSP owner

hardiepiennar:
As discussed, this keeps receiver authentication and single ownership together.
Parser/connection limits and UIBC authentication are left for separate changes;
UIBC is not removed.

The RTSP server now checks incoming clients against the selected receiver's
exact P2P group before creating a handler. The group comes from the active
NetworkManager connection or the direct supplicant connection, not from a
search across every P2P interface. An exact discovery MAC match is accepted;
for receivers with a different group MAC, the fallback requires one unambiguous
complete neighbour on that same group.

Passive connections and the active probe share a lock-protected ownership
claim. The probe can replace an unconfirmed passive claim, but not an
established owner, and a stale handler cannot release a newer claim.

I rebased this onto dev at `465d8536` and preserved the new IWD path, including
its different data-interface names.

Testing on this commit:

- 196 unit tests pass; pristine dev passes 155. Both show the same existing
  unawaited portal-cleanup coroutine warning.
- Local RTSP/FFmpeg/RTP tests pass with both supplicant-style and IWD-style
  interface fixtures: two connect/disconnect cycles, decoded video, rejection
  of unrelated and duplicate connections while media continues, and cleanup.
  Only the neighbour-command output is mocked. These are not IWD hardware tests.

Hardware comparison was repeated after the rebase on `ae1d8d07` against
pristine dev `465d8536`. With the Samsung playing a generated pattern,
a second laptop on the same Wi-Fi received M1 OPTIONS on all three unpatched
LAN connections. The patched version closed all three without sending RTSP
data, and the picture continued normally. Both runs had a successful LAN
reachability preflight and the same source-restricted firewall exception.
Normal cleanup removed the P2P link, media processes and temporary permissions.
This proves the admission difference, not full media interception.

The final cleanup shortened docstrings, restored upstream's direct interface
argument access, adjusted formatting and removed one duplicate test. Unit and
local media checks were rerun afterwards; the TV test was not repeated.

Hardware testing covers a short passive-Samsung picture/LAN-rejection test,
not a long soak. IWD, active-receiver, randomized-MAC and direct-supplicant
hardware paths remain untested.

This uses P2P-session and neighbour-table attribution, not cryptographic
receiver authentication.


IlyaP358:
Hi @hardiepiennar thanks for this. It reads as carefully as the report did, and I went through it properly rather than taking it on trust.

What holds, first. The randomized-MAC case from #44, where an LG changes its address between discovery and the group, still passes through the single-unambiguous-neighbour rule, and that was the regression I was most worried about. The active probe releases its ownership claim on both exit paths, so a failed probe cannot lock a receiver out permanently, and for sink-only receivers it never reaches the claim at all because the outbound connect is refused first. DLNA and Cast are untouched. 196 tests here against 155 on dev.

Two things before I merge, and they share a root: everything now depends on the group interface being published.

I could not reproduce this on hardware, so treat it as a mechanism rather than a sighting. Simulating an activated connection whose group interface cannot be identified, `_wait_for_nm_activation` waits out the full timeout and then reports `Timed out waiting for NetworkManager Wi-Fi Direct activation. Last status: activated`, which contradicts itself. If the interface cannot be determined I would rather it said so. Three consequences hang off that one condition, admission, activation and the probe's address lookup, so the first bug report from a setup neither of us runs would go looking in an expensive wrong place.

The second is timing symmetry. The interface race gets a two second wait, which is right. The neighbour lookup gets a single query, and the listener comes up before the link by design, so a receiver that connects before the kernel has populated the neighbour table is closed out with no second attempt. The same bounded wait keeps the fail-closed behaviour and removes the race.

On coverage, and this is not a criticism since you said it yourself, between us this has run on one television and one backend. The change turns a wrong answer into a refused connection rather than a degraded one, so the failure mode is total for whoever hits it. I am out of time today, but I will test it against my Samsung tomorrow, and once it is in `dev` I will ask the regulars in our Discord to run it on other hardware, LG webOS and a projector at least, before it goes anywhere near a release. That is how the UIBC work and the LG debugging landed here.

I will get to #144 tomorrow as well. Since it currently contains this branch, I would like to take them in order rather than in parallel. I have a couple of smaller notes on the server and handler, but they sit in the files #144 reworks, so they belong on that review rather than this one.

hardiepiennar:
Thanks Illia. 

Both points are addressed in 377d871a697feedf742db2f5138d48b26be5b0f6. The activation error now identifies the missing group interface, and neighbor verification retries within a two-second budget, releasing the lookup lock between attempts.

All 213 tests and the local streaming checks pass. I haven’t rerun hardware on this follow-up.
I’ve left #144 unchanged so we can take them in order as you suggested.

Ready for another look when you have time.

IlyaP358 (approved):
Nice. Both fixes are right, the activation error now says what actually happened instead of contradicting itself, and the retry loop shares one deadline, releases the lock before sleeping and bounds the lookup itself. A legitimate receiver still costs nothing measurable, a rejection costs the full two seconds. 213 tests here too.

My Samsung on Hyprland through NetworkManager, twice, connected and streamed both times.

Merging. Like i said i will ask the regulars with other hardware to run `dev` before this goes into a release, and I will start on #144 next.

PR #144, opened 2026-09-20
Bound RTSP messages and connection resources

hardiepiennar:
Follow-up to #143 for the parser and connection-limits advisory. This branch currently includes #143; the limits change is 54a8235a65fdd4601ca92b9ad04667ff0cd8bb04.

Both passive RTSP and the active probe now use bounded message reads: 8 KiB per line, 64 headers, 32 KiB of headers in total, and a 64 KiB body. Content-Length and media numeric fields are checked before conversion. Matching duplicate headers are accepted, including the duplicate Content-Length seen from Samsung; conflicting values and truncated messages are rejected.

An incomplete message has a 30-second deadline which incoming bytes cannot extend. Idle negotiation is bounded too, but silence during PLAY does not impose a cast-duration limit. Socket writes have an 8-second timeout. At most four RTSP connections are tracked, including the active probe, and Stop interrupts blocked readers.

Pending requests and keepalive/health-probe timer chains are also bounded. Older unanswered keepalives are discarded without stopping a healthy stream.

Verification on this commit: 228 unit tests pass, including slow-drip input, floods, numeric bounds, connection-slot recovery and timer cleanup. A safe control comparison against #143 reproduces the missing header/body bounds. Local TCP/FFmpeg/RTP tests pass negotiation, decoded video, reconnect, TEARDOWN, EOF and server Stop; a silent PLAY session continues beyond the 30-second message deadline.

Physical Samsung testing on this commit passes first connection and reconnect, user-confirmed moving-picture playback, keepalive responses and planned-stop cleanup. Healthy media was observed beyond 65 and 80 seconds after PLAY respectively. An earlier attempt timed out before RTSP after a missed TV approval prompt and needed removal of its unused P2P interface; that failed-start cleanup issue remains separate from this change. UIBC authentication is unchanged and remains separate.


IlyaP358:
@hardiepiennar, #143 is in `dev`, thank you. This branch conflicts now, but not for the reason it looks like, so here is the shortcut.

I squash-merged #143, so its individual commits are not ancestors of `dev`. This branch also sits on `28a338e`, which is #143 before the two follow-ups you pushed yesterday. A plain rebase onto `dev` would therefore replay all of #143 on top of a `dev` that already has it, and hand you seven files of your own code to resolve by hand.

Dropping what `dev` already carries does it instead:

```bash
git rebase --onto origin/dev 28a338e security/rtsp-limits-20260920
```

I tried it here: clean, no conflicts, one commit left, and 245 tests pass afterwards. So it should just be a push.

Once it is up I will start reviewing straight away =]

hardiepiennar:
Thanks Illia, rebased and pushed. :-D

IlyaP358 (approved):
Tested on my Samsung, connected and streamed. Merging.

Thank you, genuinely. You came with a private report and then did the work yourself, twice over, and took the review notes without a fuss. That is not how most security reports go, and it helped me enormously.

You have a place on the [contributors page](https://fluxcast.dev/contributors.html) like everyone else who has left a mark here.

Both fixes ship in a release today.

